Issue #12, opened 2020-09-15
Plugin system for code formatting


hukkinj1:
Make a plugin system for code formatting.

As an example, we should be able to `pip install mdformat-black` after which all Python code blocks will be formatted with Black.

E.g.
```markdown
~~~python
'''black hates single quotes'''
~~~
```
will be formatted as 
```markdown
~~~python
"""black hates single quotes"""
~~~
```

PR #15, opened 2020-09-17
Plugin system for code formatting #12

hukkinj1:
Closes https://github.com/executablebooks/mdformat/issues/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ mdformat.file("README.md")
 
 # ...or a pathlib.Path object
 import pathlib
+
 filepath = pathlib.Path("README.md")
 mdformat.file(filepath)
 ~~~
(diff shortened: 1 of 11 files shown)


mdformat/_api.py
```diff
@@ -1,17 +1,22 @@
...
 
 
-def text(md: str) -> str:
+def text(md: str, *, codeformatters: Iterable[str] = ()) -> str:
     """Format a Markdown string."""
-    return MarkdownIt(renderer_cls=MDRenderer).render(md)
+    markdown_it = MarkdownIt(renderer_cls=MDRenderer)
+    markdown_it.options["codeformatters"] = {
+        lang: mdformat.plugins.CODEFORMATTERS[lang] for lang in codeformatters
+    }
```
hukkinj1, line 15: @chrisjsewell I want to pass config to `MDRenderer` here. More specifically I need to give it a list of coding languages that I want to enable formatting for.

Question is, is this the way way to pass config to the renderer, or is there a more idiomatic way? Other ways I could think of were
- `MDRenderer.__init__` (but markdown-it-py wont let me change the args)
- Config arg of `MarkdownIt` (but that overrides all config, I just want to add this one dict in)
- `MarkdownIt.configure()` and `MarkdownIt.configure()` (i guess the same override problem here?)

chrisjsewell: Yes I would say that this is the correct way, and how I do it in myst-parser 👍

README.md
```diff
@@ -108,3 +109,22 @@ Add the following to your project's `.pre-commit-config.yaml` to enable this:
   hooks:
   - id: mdformat
 ~~~
+
+## Code formatter plugins
+
+Mdformat features a plugin system to support formatting of Markdown code blocks where the coding language has been labeled.
+For instance, if [`mdformat-black`](<https://github.com/hukkinj1/mdformat-black>) plugin is installed in the environment,
```
chrisjsewell, line 116: perhaps a note/example about how you can achieve this with your own plugin; by adding and an `mdformat.codeformatter` entry point https://packaging.python.org/guides/distributing-packages-using-setuptools/#entry-points

hukkinj1: Thank you for the review! Yeah makes sense to add this. I'm thinking to maybe add a CONTRIBUTING.md and add this there, in order to not mix up docs intended for users versus devs.

chrisjsewell: Yeh no problem. FYI, since there is currently no CONTRIBUTING.md, the contributing guidelines are directed towards https://github.com/executablebooks/.github/blob/master/CONTRIBUTING.md (see for example the bottom of this page). Since they will now be directed to the repo one, you might want to give them a mention in your CONTRIBUTING.md.

If you didn't notice as well, things like the issue templates and code of conduct are inherited by default from this repo for all others in the organisation.

PR #17, opened 2020-09-17
Bump version: 0.1.2 → 0.1.3

hukkinj1:


chrisjsewell:
@hukkinj1 That was a quick turn around in version updates lol.
BTW, have you considered adding the pypi release as a GH Actions job: like https://github.com/executablebooks/markdown-it-py/blob/00df400cfe100a962a2eb8c30e44f65ebc7f82a9/.github/workflows/tests.yml#L82

hukkinj1:
Yeah I should definitely configure the release job when I can find the time.

Regarding the two version updates, the commit that I tagged with 0.1.2 (https://github.com/executablebooks/mdformat/pull/16/commits) did not end up in master branch. The commit hash is in master is different, I think because I did a rebase? Did a merge with 0.1.3 so all seems good there. I'm used to just pushing tagged commits to master so this process is still a bit new to me...

chrisjsewell:
Oh you tag before committing, yeh rebasing makes new commits so changed the hashes. No right way really, but yeh I just commit then tag on master, which then triggers the publish workflow

hukkinj1:
:+1: I've been using https://github.com/c4urself/bump2version to make a version commit and associated tag. It should be possible to configure publish workflow to trigger when a tagged commit like this is merged to master.

chrisjsewell:
Yeh I noticed that, sounds like it should be possible 👍

PR #29, opened 2020-09-24
✨ Add plugin system for extending the parser

### LINKED ISSUE #13: Plugin system for extending the parser
Perhaps similar to #12, it would be nice to provide a hook for extending the parser with the available plugins: https://markdown-it-py.readthedocs.io/en/latest/plugins.html#plugin-extensions

As mentioned in executablebooks/markdown-it-py#10, I guess this would work best such that it is the plugin's responsibility to provide a parser method? Maybe allowing for the parsing of one or more renderer defined options, something like:

```python
md = MarkdownIt(renderer_cls=MDRenderer)
md.use(a_plugin, md_seperator=MARKERS.BLOCK_SEPARATOR)
md.render(md)
```

chrisjsewell:
~~This is not quite the finished product yet, but it outlines the basic functionality of what I'm looking to implement~~ now basically complete

closes #13

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,8 +16,26 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.7
     - uses: pre-commit/action@v2.0.0
+
+  mypy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+
+    - name: Installation (deps and package)
+      run: |
+        pip install "poetry==1.0.10"
+        poetry config virtualenvs.create false && poetry install --no-interaction --no-ansi
+
+    - name: run mypy
+      run: mypy .
 
   tests:
     runs-on: ${{ matrix.os }}
@@ -48,23 +66,19 @@ jobs:
       uses: codecov/codecov-action@v1.0.12
 
   pre-commit-hook:
+
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [3.7]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
+    - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: 3.7
 
     - name: Install pre-commit
       run: |
-        python -m pip install --upgrade pip
         pip install pre-commit
 
     - name: run hook against repository
       run: pre-commit try-repo . mdformat --verbose --all-files --show-diff-on-failure
-
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,6 @@ on:
 jobs:
 
   pre-commit:
-
     runs-on: ubuntu-latest
 
     steps:
@@ -20,7 +19,7 @@ jobs:
         python-version: 3.6
     - uses: pre-commit/action@v2.0.0
 
-  build:
+  tests:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
@@ -40,11 +39,6 @@ jobs:
         pip install "poetry==1.0.10"
         poetry config virtualenvs.create false && poetry install --no-interaction --no-ansi
 
-    - name: Run formatter on repo files
-      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.6'
-      run: |
-        mdformat --check CHANGELOG.md CONTRIBUTING.md README.md
-
     - name: Test with pytest
       run: |
         pytest --cov --cov-fail-under=95
@@ -52,3 +46,25 @@ jobs:
     - name: Report coverage
       if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.6'
       uses: codecov/codecov-action@v1.0.12
+
+  pre-commit-hook:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install pre-commit
+      run: |
+        python -m pip install --upgrade pip
+        pip install pre-commit
+
+    - name: run hook against repository
+      run: pre-commit try-repo . mdformat --verbose --all-files --show-diff-on-failure
+
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,6 +7,19 @@ on:
     branches: [ master ]
 
 jobs:
+
+  pre-commit:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
+    - uses: pre-commit/action@v2.0.0
+
   build:
     runs-on: ${{ matrix.os }}
     strategy:
@@ -27,14 +40,9 @@ jobs:
         pip install "poetry==1.0.10"
         poetry config virtualenvs.create false && poetry install --no-interaction --no-ansi
 
-    - name: Linters
+    - name: Run formatter on repo files
       if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.6'
       run: |
-        flake8 .
-        mypy .
-        isort --check-only .
-        black --check .
-        docformatter --recursive --check mdformat/ tests/
         mdformat --check CHANGELOG.md CONTRIBUTING.md README.md
 
     - name: Test with pytest
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -27,7 +27,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
         return 1
 
     # Enable all parser plugins
-    enabled_parserplugins_langs = mdformat.plugins.EXTENDPLUGINS.keys()
+    enabled_parserplugins = mdformat.plugins.PARSER_EXTENSIONS.keys()
     # Enable code formatting for all languages that have a plugin installed
     enabled_codeformatter_langs = mdformat.plugins.CODEFORMATTERS.keys()
 
@@ -41,7 +41,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
             original_str = sys.stdin.read()
         formatted_str = mdformat.text(
             original_str,
-            plugins=enabled_parserplugins_langs,
+            extensions=enabled_parserplugins,
             codeformatters=enabled_codeformatter_langs,
         )
 
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -27,7 +27,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
         return 1
 
     # Enable all parser plugins
-    enabled_parserplugins_langs = mdformat.plugins.PARSEPLUGINS.keys()
+    enabled_parserplugins_langs = mdformat.plugins.EXTENDPLUGINS.keys()
     # Enable code formatting for all languages that have a plugin installed
     enabled_codeformatter_langs = mdformat.plugins.CODEFORMATTERS.keys()
 
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -26,6 +26,8 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
         sys.stderr.write(f'Error: File "{e.path}" does not exist.\n')
         return 1
 
+    # Enable all parser plugins
+    enabled_parserplugins_langs = mdformat.plugins.PARSEPLUGINS.keys()
     # Enable code formatting for all languages that have a plugin installed
     enabled_codeformatter_langs = mdformat.plugins.CODEFORMATTERS.keys()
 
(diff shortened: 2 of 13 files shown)


chrisjsewell:
On a side note, it might be better to run a separate linting job, directly using pre-commit, rather than running all the tools separately?

```yaml
  pre-commit:

    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2
    - name: Set up Python 3.8
      uses: actions/setup-python@v1
      with:
        python-version: 3.8
    - uses: pre-commit/action@v2.0.0
```

chrisjsewell:
yeh this linting error doesn't even show up in pre-commit so a bit difficult to fix 🤷

chrisjsewell:
Right, rebased and now that's a better CI set up IMO 😬

chrisjsewell:
so the current code will handle "single token" type plugins, but I'm not sure what the best route is for container type plugins.
The most important one to my mind is tables, which are pretty much enabled in every markdown renderer.

chrisjsewell:
tables are a real tricky one, because I think to do it correctly you need to first generate the text for all cells, then use that to work out the width of each column (i.e. the max length of text), then write the whole table (adding correct padding). Something like:

```
| a | b |
| :-: | -: |
| cde | fghi | 
```

to 

```python
align = ["center", "right"]
cells = [["a", "b"], ["cde", "fghi"]]
```

to

```
|  a  |    b |
| :-: | ---: |
| cde | fghi |
```

chrisjsewell:
I think here then you need to be able to do two things:

1. parse the render method to the plugin, so that it can do its own nested rendering
2. allow the plugin to also return the final token index "consumed" (e.g. `table_close`), which would require something like:

rather than using:

```python
for i, token in enumerate(tokens):
```

use: 

```python
i = 0
while i < len(tokens):
  token = tokens[i]
  ...
  i += 1
```

chrisjsewell:
OK, in 8a27caa I have implemented the changes similar to those described above.
This should now allow for the accommodation of any type of extension plugin.

chrisjsewell:
I've also made `MDRenderer` and `MARKERS` available at the top-level, so they can be used by these plugins

mdformat/_cli.py
```diff
@@ -26,6 +26,8 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
         sys.stderr.write(f'Error: File "{e.path}" does not exist.\n')
         return 1
 
+    # Enable all parser plugins
+    enabled_parserplugins_langs = mdformat.plugins.EXTENDPLUGINS.keys()
```
hukkinj1, line 30: Rename as `enabled_parserplugins`?

mdformat/plugins.py
```diff
@@ -15,3 +20,33 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
 
 
 CODEFORMATTERS: Mapping[str, Callable[[str, str], str]] = _load_codeformatters()
+
+
+class ParsePluginAbstract:
```
hukkinj1, line 25: Is it intended that plugins inherit from this? Maybe make this a `abc.ABC` and make the methods `@abstractmethod`s.

mdformat/plugins.py
```diff
@@ -15,3 +20,33 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+    """A class for extending the base parser."""
+
+    @staticmethod
+    def update_mdit(mdit: MarkdownIt) -> MarkdownIt:
```
hukkinj1, line 29: I would make this return `None` to make it absolutely clear that this modifies the argument `mdit` in place and does not (and should not) instantiate anything new.

mdformat/plugins.py
```diff
@@ -15,3 +20,33 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+        """Update the parser, e.g. by adding a plugin: `mdit.use(myplugin)`"""
+        return mdit
+
+    @staticmethod
+    def render_token(
+        renderer: MDRenderer, tokens: List[Token], index: int, options: dict, env: dict
```
hukkinj1, line 35: Is there a use case for the `renderer` arg here?

mdformat/plugins.py
```diff
@@ -15,3 +20,33 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+    ) -> Optional[Tuple[str, int]]:
+        """Convert a token to a string, or return None if no render method
+        available.
+
+        :returns: (text, index) where index is of the final "consumed" token
+        """
+        return None
+
+
+def _load_extendplugins() -> Mapping[str, ParsePluginAbstract]:
```
hukkinj1, line 45: I would make the return type `Dict` instead of `Mapping` here.

mdformat/plugins.py
```diff
@@ -15,3 +20,33 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+    extendplugins_entrypoints = importlib_metadata.entry_points().get(
+        "mdformat.extendplugins", ()
```
hukkinj1, line 47: The group for codeformatters was made singular form `mdformat.codeformatter`, so maybe this too for consistency? Not sure if one or the other is more appropriate.

pyproject.toml
```diff
@@ -44,16 +44,7 @@ importlib-metadata = { version = ">=0.12", python = "<3.8" }
...
 pytest-cov = "*"
-
-# Linting
-isort = "*"
-black = { version = "*", allow-prereleases = true }
-flake8 = "*"
-flake8-bugbear = "*"
-flake8-builtins = "*"
-flake8-comprehensions = "*"
-mypy = "*"
```
hukkinj1, line 55: Just a note regarding the earlier comment about mypy in CI: we may have to add mypy back here.

pyproject.toml
```diff
@@ -77,3 +68,29 @@ profile = "black"
...
+
+[tool.tox]
+legacy_tox_ini = """
+[tox]
+envlist = py{36,37,38}
+isolated_build = True
+
+[testenv:py{36,37,38}]
+# from https://github.com/python-poetry/poetry/issues/1941
+# it appears there is no easy way to access [tool.poetry.dev-dependencies]
```
hukkinj1, line 80: I dont know much about tox or tox+poetry config, but maybe see https://python-poetry.org/docs/faq/#is-tox-supported

I would imagine we have to tell tox to run `poetry install` somehow?

hukkinj1: Also should add mypy here somehow perhaps?

hukkinj1 (changes requested):

hukkinj1:
Thanks for the PR! Some questions and feedback above.

I like the "return token index consumed" approach.

hukkinj1:
Oh also, maybe add a brief description how to make a parser plugin in CONTRIBUTING.md?

mdformat/plugins.py
hukkinj1: On second though, would something like `"mdformat.parser_extension`" be more descriptive? Maybe the constant below should also be `mdformat.plugins.PARSER_EXTENSIONS`. The plugin part should be implied from the namespace, i think.

hukkinj1: Oh and if we do this name change, it should probably be applied to keyword arguments of `mdformat.file()` and `mdformat.text()` as well

.github/workflows/tests.yml
```diff
@@ -27,20 +39,32 @@ jobs:
         pip install "poetry==1.0.10"
         poetry config virtualenvs.create false && poetry install --no-interaction --no-ansi
 
-    - name: Linters
-      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.6'
-      run: |
-        flake8 .
-        mypy .
```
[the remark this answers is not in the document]
chrisjsewell: ah yes I missed that one, thats a pain.
would it not be better to add mypy to pre-commit config as a local hook?
otherwise I think it should be part of either the pre-commit job, or just a seperate job, rather than mixed in to the pytest runs

.github/workflows/tests.yml
```diff
@@ -27,20 +39,32 @@ jobs:
...
 
     - name: Report coverage
       if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.6'
       uses: codecov/codecov-action@v1.0.12
+
+  pre-commit-hook:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7]
```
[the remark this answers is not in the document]
chrisjsewell: can do, if we definitely don't want to run against multiple versions?

.github/workflows/tests.yml
```diff
@@ -27,20 +39,32 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install pre-commit
+      run: |
+        python -m pip install --upgrade pip
```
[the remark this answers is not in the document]
chrisjsewell: just a copy/paste from another project really. I guess it doesn't hurt to have the latest version of pip, but probably not as important for this project, where there is not too many dependencies to resolve and we are using poetry.lock anyway.
I'm happy either way; to remove or add to `tests`?

.github/workflows/tests.yml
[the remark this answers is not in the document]
hukkinj1: > would it not be better to add mypy to pre-commit config as a local hook?

I think I would maybe prefer not to, in order to keep the hooks completely independent of any external envs.

> otherwise I think it should be part of either the pre-commit job, or just a seperate job, rather than mixed in to the pytest runs

You are probably right. I would vote for separate job.

.github/workflows/tests.yml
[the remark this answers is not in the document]
hukkinj1: IMO one version should suffice here.

.github/workflows/tests.yml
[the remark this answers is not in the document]
chrisjsewell: yeh the local hooks are a pain as obviously you can't install pre-commit, there I use tox to run pre-commit.
But yeh I will add a seperate job, and also add the running of mypy to tox

.github/workflows/tests.yml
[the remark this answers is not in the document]
chrisjsewell: 👍

.github/workflows/tests.yml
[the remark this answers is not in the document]
hukkinj1: Ok, I see. I would remove, if we don't see a strict need for it.

mdformat/_cli.py
chrisjsewell: oh yeh just a copy/paste error

mdformat/plugins.py
chrisjsewell: yeh I wasn't sure about that. I don't think its really necessary to inherit from this as it wouldn't add any extra functionality to do so. You should even actually be able to point the entry point to a module, containing the two functions (hence why they are set as staticmethods)
This was primarily for use in typing, unless there is a better to do this?

mdformat/plugins.py
chrisjsewell: yep fair, that change shouldn't anything, although just to note I was actually somewhat copying here the implementation in https://code.visualstudio.com/api/extension-guides/markdown-extension#adding-support-for-new-syntax-with-markdownit-plugins

mdformat/plugins.py
chrisjsewell: definitely; see my explanation in https://github.com/executablebooks/mdformat/pull/29#issuecomment-698216949, whereby for the full tables renderer (and likely other extensions) you will need to perform nested rendering.

mdformat/plugins.py
chrisjsewell: I copied that from `CODEFORMATTERS = ...` 😉 so I guess that should also be changed?

mdformat/plugins.py
chrisjsewell: yep sounds reasonable

pyproject.toml
chrisjsewell: 👍

pyproject.toml
chrisjsewell: yep mentioned this above.
In general though this is all working.
No you don't need to call `poetry install`, the installs are part of the build process

chrisjsewell:
> Oh also, maybe add a brief description how to make a parser plugin in CONTRIBUTING.md?

yep was just waiting to do this once the mechanism had been finalised

.github/workflows/tests.yml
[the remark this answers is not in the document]
chrisjsewell: 👍

mdformat/plugins.py
hukkinj1: I think https://docs.python.org/3/library/typing.html#typing.Protocol might be the appropriate class here, as that doesn't imply inheritance, it simply tells the type checker what interface should be expected. The annoying thing is that in Python <3.8 you would have to import that from https://pypi.org/project/typing-extensions/.

I would probably just be lazy and typehint as `typing.Any` (and remove this abstract class), but also if you want to do it, I have nothing against a `Protocol` class (in which case you have to add something like `typing-extensions= { version = "*", python = "<3.8" }` to `pyproject.toml`.

mdformat/plugins.py
hukkinj1: I see. I would still make the change, as I think that's how mutator function interfaces should be (and how e.g. `dict.update()` from stdlib behaves).

mdformat/plugins.py
hukkinj1: Thanks, got it!

mdformat/plugins.py
hukkinj1: I mean just `_load_extendplugins` here.

`_load_codeformatters` actually is already typehinted as `-> Dict[...`, as that is the accurate type it returns.

For the public constant `CODEFORMATTERS` I then set it to `Mapping` (which is immutable), to make type checker take care that nobody mutates the dict.

chrisjsewell: and I see, that makes sense

chrisjsewell:
all review comments addressed!

hukkinj1:
This is fantastic! In the words of the infinitely wise welcome-bot

> You are awesome! :hugs:

Merged.

chrisjsewell:
Ah it always brightens my day to see the party fish 🎉 🤣

hukkinj1:
@chrisjsewell I think I will make one change myself, which is expose `mdformat.renderer` package instead of `mdformat.MDRenderer` and `mdformat.MARKERS`. I think especially with the `MARKERS` it feels better to me to have it in the renderer namespace.

I will then release, shouldn't take very long.

chrisjsewell:
yeh no worries ta

PR #35, opened 2020-09-25
✨ Allow options to be updated on the `MarkdownIt` instance

chrisjsewell:
This provides a hook for passing variables to the parsing/render process,
which in-turn provides a simple way of allowing plugins to accept variable arguments from the API or CLI.

~~Note this is back-incompatible for parser plugins, since `update_mdit` now also gets passed this env dict.~~

### DIFF
--- a/mdformat/_api.py
+++ b/mdformat/_api.py
@@ -18,7 +18,7 @@ def text(
     markdown_it = MarkdownIt(renderer_cls=MDRenderer)
     # store reference labels in link/image tokens
     markdown_it.options["store_labels"] = True
-    markdown_it.options["mdformat"] = options
+    markdown_it.options["mdformat"] = options or {}
 
     markdown_it.options["parser_extension"] = []
     for name in extensions:
--- a/mdformat/_api.py
+++ b/mdformat/_api.py
@@ -2,7 +2,6 @@ from pathlib import Path
 from typing import Iterable, Optional, Union
 
 from markdown_it import MarkdownIt
-from markdown_it.utils import AttrDict
 
 import mdformat.plugins
 from mdformat.renderer import MDRenderer
@@ -11,7 +10,7 @@ from mdformat.renderer import MDRenderer
 def text(
     md: str,
     *,
-    env: Optional[dict] = None,
+    options: Optional[dict] = None,
     extensions: Iterable[str] = (),
     codeformatters: Iterable[str] = (),
 ) -> str:
@@ -19,23 +18,24 @@ def text(
     markdown_it = MarkdownIt(renderer_cls=MDRenderer)
     # store reference labels in link/image tokens
     markdown_it.options["store_labels"] = True
+    markdown_it.options.update(options or {})
 
     markdown_it.options["parser_extension"] = []
     for name in extensions:
         plugin = mdformat.plugins.PARSER_EXTENSIONS[name]
-        plugin.update_mdit(markdown_it, env)
+        plugin.update_mdit(markdown_it)
         markdown_it.options["parser_extension"].append(plugin)
     markdown_it.options["codeformatters"] = {
         lang: mdformat.plugins.CODEFORMATTERS[lang] for lang in codeformatters
     }
-    env = env if env is None else AttrDict(env)
-    return markdown_it.render(md, env=env)
+
+    return markdown_it.render(md)
 
 
 def file(
     f: Union[str, Path],
     *,
-    env: Optional[dict] = None,
+    options: Optional[dict] = None,
     extensions: Iterable[str] = (),
     codeformatters: Iterable[str] = (),
 ) -> None:
@@ -50,6 +50,9 @@ def file(
         raise ValueError(f'Can not format "{f}". It is not a file.')
     original_md = f.read_text(encoding="utf-8")
     formatted_md = text(
-        original_md, env=env, extensions=extensions, codeformatters=codeformatters
+        original_md,
+        options=options,
+        extensions=extensions,
+        codeformatters=codeformatters,
     )
     f.write_text(formatted_md, encoding="utf-8")
--- a/mdformat/_api.py
+++ b/mdformat/_api.py
@@ -1,14 +1,19 @@
 from pathlib import Path
-from typing import Iterable, Union
+from typing import Iterable, Optional, Union
 
 from markdown_it import MarkdownIt
+from markdown_it.utils import AttrDict
 
 import mdformat.plugins
 from mdformat.renderer import MDRenderer
 
 
 def text(
-    md: str, *, extensions: Iterable[str] = (), codeformatters: Iterable[str] = ()
+    md: str,
+    *,
+    env: Optional[dict] = None,
+    extensions: Iterable[str] = (),
+    codeformatters: Iterable[str] = (),
 ) -> str:
     """Format a Markdown string."""
     markdown_it = MarkdownIt(renderer_cls=MDRenderer)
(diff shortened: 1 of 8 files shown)


hukkinj1:
Hey, could you help me understand a few things.

We already pass some state to the parser and render in the `options` dict. Why not use that here as well? I'm having a bit of a hard time understanding the `options` and `env` distinction. I've only seen the `env` used for reference link refs, so thought that the use-case for that is parser output that for one reason or the other doesn't fit in the token stream. The definition [here](https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md):
> An env sandbox can be used alongside tokens to inject external variables for your parsers and renderers.

Makes me think that the source of `env` data is not always the parser. Then how is it different from `options`?

hukkinj1:
Also would be interested in how you plan to use this feature exactly? Is there a use case for `update_mdit` receiving the `env` or is it enough if the render function does so, etc?

chrisjsewell:
> I'm having a bit of a hard time understanding the options and env distinction in the markdown-it world in general

The difference is that `options` should remain constant over multiple text parses, whereas `env` is transient and gets reset for each parse.
For example, you should set up your `MarkdownIt` object once then reuse it for all parsing:

```python
mdit = MarkdownIt()
# this option wil be used for every parse
mdit.option["a"] = 1
# but the env will only be used for the parse it is set on
mdit.parse("hallo", env={"b": 1})
mdit.parse("hallo")
```

So yeh you are probably right that this data should update options.

Note, mdformat anyhow diminishes this distinction a bit, because a new `MarkdownIt` instance is created for every parse.
I guess technically, in `mdformat._cli.run`, you should set up a single `MarkdownIt` instance and use it to parse all files, rather than calling `mdformat.text` and creating a new one every time.
(this is more important in long running/responsive programs, for example VS Code sets up a single `MarkdownIt` instance in memory, and will use it for every call to update a preview)

> Also would be interested in how you plan to use this feature exactly?

Any way I want lol.
If a plugin wants to be a bit more "unopinionated" (i.e. accepting configuration) then this should be allowed.
For an example though, `markdown_it.extensions.texmath.texmath_plugin` can be initialized with different styles of delimiter: `dollars`, `brackets`, `gitlab`, etc.
I don't want to have to create a separate plugin for every style, over just calling something like  `mdformat -e texmath-delim=brackets`

chrisjsewell:
In 0672535 I've moved to setting options rather than env

hukkinj1:
Thanks for the explanation!

> ...how you plan to use this feature exactly?
> > Any way I want lol...

Aha, if it's as general purpose as this, then what do you think about the following: we give plugins the possibility to add `argparse` arguments of their own. Then in the render phase we'd pass all CLI args to the plugin. This way they the plugins can add their own `--help` strings to document the argument, argparse will raise if two plugins use the same name, etc.

I think we'd need another function to the plugin interface, something like
```python
def add_options(parser: argparse.ArgumentParser) -> None:
    ...
```
which is where a plugin would add its options.

Then in `mdformat._cli.run` we could pass a mapping of the parsed args to the formatter function `mdformat.text` to make the options usable from Python API as well as the CLI:
```python
import mdformat
options = vars(parser.parse_args(cli_args))
mdformat.text("bla bla", options=options)
```

hukkinj1:
We could make defining `add_options` optional for the plugin by getting the function from the plugin with `getattr` and some default.

What do you think?

chrisjsewell:
> What do you think?

Yeh absolutely, I thought you'd be more willing to accept a simpler approach lol, but that is more of a "complete" solution 👍

chrisjsewell:
> Then in the render phase we'd pass all CLI args to the plugin

I think these options should still be stored in `mdit.options` though, potentially under a top-level key, e.g.

```python
mdit = MarkdownIt()
options = vars(parser.parse_args(cli_args))
mdit.options["mdformat"] = options
```

chrisjsewell:
> We could make defining add_options optional for the plugin by getting the function from the plugin with getattr and some default.

Yeh thts what I have done in #36, for allowing plugins to specify certain HTML classes to ignore when comparing the input/output HTML

hukkinj1:
> potentially under a top-level key

I would use a top-level key too. Not sure, but perhaps even two keys, if we want to fit the current "codeformatters" and "parser_extensions" in the same top-level key. Something like:

```python
options["mdformat"]["opts"]
options["mdformat"]["codeformatters"]
options["mdformat"]["parser_extensions"]
```

mdformat/_cli.py
```diff
@@ -23,9 +29,11 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
     try:
         file_paths = resolve_file_paths(args.paths)
     except InvalidPath as e:
-        sys.stderr.write(f'Error: File "{e.path}" does not exist.\n')
-        return 1
+        parser.error(f'File "{e.path}" does not exist.\n')
```
hukkinj1, line 32: Not sure but I would guess the newline at the end here is not needed now? (I.e. `parser.error` would act more like a `print()` than a `write()`)

mdformat/_cli.py
```diff
@@ -23,9 +29,11 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
...
 
+    # convert args to dict
+    options = dict(args._get_kwargs())
```
hukkinj1, line 35: Could we use `vars(args)` here? I wouldn't want to use private API.

mdformat/_cli.py
```diff
@@ -23,9 +29,11 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
...
+    options.pop("paths")
```
hukkinj1, line 36: I suggest we remove this and just pass all args to the plugins. The logic is simpler that way.

mdformat/plugins.py
```diff
@@ -27,6 +28,11 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
 class ParserExtensionInterface(Protocol):
     """A interface for parser extension plugins."""
 
+    def add_cli_options(self, parser: argparse.ArgumentParser) -> None:
+        """Add options to the mdformat CLI, to be stored in
+        mdit.options["mdformat"]"""
```
hukkinj1, line 33: Maybe a mention that defining this function is optional?

mdformat/plugins.py
```diff
@@ -27,6 +28,11 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+        pass
```
hukkinj1, line 34: The `pass` should be redundant here, since the docstring is a Python statement. (In fact same applies to all method implementations in this class)

hukkinj1 (changes requested):

mdformat/_api.py
```diff
@@ -8,12 +8,17 @@
...
+    *,
+    options: Optional[dict] = None,
+    extensions: Iterable[str] = (),
+    codeformatters: Iterable[str] = (),
 ) -> str:
     """Format a Markdown string."""
     markdown_it = MarkdownIt(renderer_cls=MDRenderer)
     # store reference labels in link/image tokens
     markdown_it.options["store_labels"] = True
+    markdown_it.options["mdformat"] = options
```
chrisjsewell, line 21:
```suggestion
    markdown_it.options["mdformat"] = options or {}
```

mdformat/_api.py
```diff
@@ -8,12 +8,17 @@
...
 def text(
-    md: str, *, extensions: Iterable[str] = (), codeformatters: Iterable[str] = ()
+    md: str,
+    *,
+    options: Optional[dict] = None,
+    extensions: Iterable[str] = (),
+    codeformatters: Iterable[str] = (),
 ) -> str:
     """Format a Markdown string."""
     markdown_it = MarkdownIt(renderer_cls=MDRenderer)
```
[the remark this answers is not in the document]
chrisjsewell: done below

mdformat/_cli.py
chrisjsewell, line 32:
```suggestion
        parser.error(f'File "{e.path}" does not exist.')
```

mdformat/_cli.py
chrisjsewell, line 35:
```suggestion
    options = vars(args)
```

mdformat/plugins.py
chrisjsewell, line 33:
```suggestion
        mdit.options["mdformat"] (optional)"""
```

mdformat/plugins.py
chrisjsewell, line 34:
```suggestion
```

mdformat/plugins.py
```diff
@@ -27,6 +28,11 @@ def _load_codeformatters() -> Dict[str, Callable[[str, str], str]]:
...
+
     def update_mdit(self, mdit: MarkdownIt) -> None:
         """Update the parser, e.g. by adding a plugin: `mdit.use(myplugin)`"""
         pass
```
chrisjsewell, line 38:
```suggestion
```

hukkinj1:
Looking very good! There's one unresolved comment I think you might have missed.

chrisjsewell:
yeh I just had to fix the test

hukkinj1:
Beautiful, thanks! Merged

PR #36, opened 2020-09-26
👌 Improve MD comparison and testing

chrisjsewell:
1. Add plugin extensions to renderer before generating HTML 
2. Use HTML parsing to compare outputs; I think this is a bit more of an elegant solution
3. Use the markdown-it-py fixtures approach to testing input vs expected output rendering 😄 Its easier to add/modify tests, and serves as nice "specification" for the renderer that people can look at to quickly understand its behaviour (which is not as clear when looking at a JSON blob)
4. Add some extra tests for code formatter plugins

### DIFF
--- a/mdformat/_util.py
+++ b/mdformat/_util.py
@@ -31,7 +31,7 @@ def is_md_equal(
 
 
 class HTML2AST(HTMLParser):
-    """Parser HTML to JSON."""
+    """Parser HTML to AST."""
 
     def parse(self, text: str, strip_classes: Iterable[str] = ()) -> List[dict]:
         self.tree: List[dict] = []
@@ -72,7 +72,9 @@ class HTML2AST(HTMLParser):
             self.current = self.current.pop("parent")
 
     def handle_data(self, data: str) -> None:
-        # ignore empty data
-        if data.strip():
-            assert self.current is not None, f"'{data}'"
-            self.current["data"] = data.rstrip()
+        # ignore data outside tabs
+        if self.current is not None:
+            # ignore empty lines and trailing whitespace
+            self.current["data"] = [
+                li.rstrip() for li in data.splitlines() if li.strip()
+            ]
--- a/mdformat/_util.py
+++ b/mdformat/_util.py
@@ -1,21 +1,81 @@
-import re
-from typing import Iterable
+from html.parser import HTMLParser
+from typing import Iterable, List, Optional, Set
 
 from markdown_it import MarkdownIt
 
+import mdformat.plugins
 
-def is_md_equal(md1: str, md2: str, *, ignore_codeclasses: Iterable[str] = ()) -> bool:
+
+def is_md_equal(
+    md1: str,
+    md2: str,
+    *,
+    enabled_extensions: Iterable[str] = (),
+    enabled_codeformatters: Iterable[str] = (),
+) -> bool:
     """Check if two Markdown produce the same HTML.
 
-    Renders HTML from both Markdown strings, strips whitespace and
-    checks equality. Note that this is not a perfect solution, as there
-    can be meaningful whitespace in HTML, e.g. in a <code> block.
+    Renders HTML from both Markdown strings, strip content of tags with
+    specified classes, and checks equality.
     """
-    html1 = MarkdownIt().render(md1)
-    html2 = MarkdownIt().render(md2)
-    html1 = re.sub(r"\s+", "", html1)
-    html2 = re.sub(r"\s+", "", html2)
-    for codeclass in ignore_codeclasses:
-        html1 = re.sub(rf'<codeclass="language-{codeclass}">.*</pre>', "", html1)
-        html2 = re.sub(rf'<codeclass="language-{codeclass}">.*</pre>', "", html2)
+    ignore_classes = [f"language-{lang}" for lang in enabled_codeformatters]
+    mdit = MarkdownIt()
+    for name in enabled_extensions:
+        plugin = mdformat.plugins.PARSER_EXTENSIONS[name]
+        plugin.update_mdit(mdit)
+        ignore_classes.extend(getattr(plugin, "ignore_classes", []))
+    html1 = HTML2JSON().parse(mdit.render(md1), ignore_classes)
+    html2 = HTML2JSON().parse(mdit.render(md2), ignore_classes)
+
     return html1 == html2
+
+
+class HTML2JSON(HTMLParser):
+    """Parser HTML to JSON."""
+
+    def parse(self, text: str, strip_classes: Iterable[str] = ()) -> List[dict]:
+        self.tree: List[dict] = []
+        self.current: Optional[dict] = None
+        self.feed(text)
+        self.strip_classes(self.tree, set(strip_classes))
+        return self.tree
+
+    def strip_classes(self, tree: List[dict], classes: Set[str]) -> List[dict]:
+        """Strip content from tags with certain classes."""
+        items = []
+        for item in tree:
+            if set(item["attrs"].get("class", "").split()).intersection(classes):
+                items.append({"tag": item["tag"], "attrs": item["attrs"]})
+                continue
+            items.append(item)
+            item["children"] = self.strip_classes(item.get("children", []), classes)
+            if not item["children"]:
+                item.pop("children")
+
+        return items
+
+    def handle_starttag(self, tag: str, attrs: list) -> None:
+        tag_item = {"tag": tag, "attrs": dict(attrs), "parent": self.current}
+        if self.current is None:
+            self.tree.append(tag_item)
+        else:
+            children = self.current.setdefault("children", [])
+            children.append(tag_item)
+        self.current = tag_item
+
+    def handle_endtag(self, tag: str) -> None:
+        # walk up the tree to the tag's parent
+        while self.current is not None:
+            if self.current["tag"] == tag:
+                self.current = self.current.pop("parent")
+                break
+            self.current = self.current.pop("parent")
+
+    def handle_data(self, data: str) -> None:
+        if self.current:
+            self.current["data"] = data.rstrip()
+
+    def handle_comment(self, data: str) -> None:
+        if self.current:
+            self.current.setdefault("comments", [])
+            self.current["comments"].append(data)
--- a/mdformat/_util.py
+++ b/mdformat/_util.py
@@ -16,7 +16,7 @@ def is_md_equal(
     """Check if two Markdown produce the same HTML.
 
     Renders HTML from both Markdown strings, strip content of tags with
-    specified classes, and checks equality.
+    specified classes, and checks equality of the generated ASTs.
     """
     ignore_classes = [f"language-{lang}" for lang in enabled_codeformatters]
     mdit = MarkdownIt()
@@ -24,13 +24,13 @@ def is_md_equal(
         plugin = mdformat.plugins.PARSER_EXTENSIONS[name]
         plugin.update_mdit(mdit)
         ignore_classes.extend(getattr(plugin, "ignore_classes", []))
-    html1 = HTML2JSON().parse(mdit.render(md1), ignore_classes)
-    html2 = HTML2JSON().parse(mdit.render(md2), ignore_classes)
+    html1 = HTML2AST().parse(mdit.render(md1), ignore_classes)
+    html2 = HTML2AST().parse(mdit.render(md2), ignore_classes)
 
     return html1 == html2
 
 
-class HTML2JSON(HTMLParser):
+class HTML2AST(HTMLParser):
     """Parser HTML to JSON."""
 
     def parse(self, text: str, strip_classes: Iterable[str] = ()) -> List[dict]:
@@ -72,10 +72,7 @@ class HTML2JSON(HTMLParser):
             self.current = self.current.pop("parent")
 
     def handle_data(self, data: str) -> None:
-        if self.current:
+        # ignore empty data
+        if data.strip():
+            assert self.current is not None, f"'{data}'"
             self.current["data"] = data.rstrip()
-
-    def handle_comment(self, data: str) -> None:
-        if self.current:
-            self.current.setdefault("comments", [])
-            self.current["comments"].append(data)
--- a/mdformat/renderer/_token_renderers.py
+++ b/mdformat/renderer/_token_renderers.py
@@ -1,5 +1,6 @@
 """A namespace for functions that render the Markdown of tokens from markdown-
 it-py."""
+import logging
 from typing import List, Optional
 
 from markdown_it.token import Token
@@ -11,6 +12,8 @@ from mdformat.renderer._util import (
     is_text_inside_autolink,
     longest_consecutive_sequence,
 )
+
+LOGGER = logging.getLogger(__name__)
 
 
 def default(tokens: List[Token], idx: int, options: dict, env: dict) -> str:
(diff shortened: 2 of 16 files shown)


mdformat/_util.py
```diff
@@ -1,21 +1,81 @@
...
+        plugin = mdformat.plugins.PARSER_EXTENSIONS[name]
+        plugin.update_mdit(mdit)
+        ignore_classes.extend(getattr(plugin, "ignore_classes", []))
+    html1 = HTML2JSON().parse(mdit.render(md1), ignore_classes)
+    html2 = HTML2JSON().parse(mdit.render(md2), ignore_classes)
+
     return html1 == html2
+
+
+class HTML2JSON(HTMLParser):
```
hukkinj1, line 33: This seems to output a syntax tree, not JSON. Maybe rename to `HTML2AST` or similar?

mdformat/renderer/_token_renderers.py
```diff
@@ -101,9 +104,10 @@ def fence(tokens: List[Token], idx: int, options: dict, env: dict) -> str:
         fmt_func = options["codeformatters"][lang]
         try:
             code_block = fmt_func(code_block, info_str)
-        except Exception:
+        except Exception as err:
             # Swallow exceptions so that formatter errors (e.g. due to
             # invalid code) do not crash mdformat.
+            LOGGER.warning(f"Code formatting of '{lang}' failed: {err}")
             pass
```
hukkinj1, line 111: Remove this line?

mdformat/_util.py
```diff
@@ -1,21 +1,81 @@
...
+            if self.current["tag"] == tag:
+                self.current = self.current.pop("parent")
+                break
+            self.current = self.current.pop("parent")
+
+    def handle_data(self, data: str) -> None:
+        if self.current:
+            self.current["data"] = data.rstrip()
+
+    def handle_comment(self, data: str) -> None:
```
hukkinj1, line 78: Do we need to define this? Maybe just skip comments?

mdformat/_util.py
```diff
@@ -1,21 +1,81 @@
...
+    def handle_endtag(self, tag: str) -> None:
+        # walk up the tree to the tag's parent
+        while self.current is not None:
+            if self.current["tag"] == tag:
+                self.current = self.current.pop("parent")
+                break
+            self.current = self.current.pop("parent")
+
+    def handle_data(self, data: str) -> None:
+        if self.current:
```
hukkinj1, line 75: Can `self.current` ever be falsy, given valid HTML? If not, maybe remove the check and fail fast and hard?

mdformat/_util.py
```diff
@@ -1,21 +1,81 @@
...
-    html2 = re.sub(r"\s+", "", html2)
-    for codeclass in ignore_codeclasses:
-        html1 = re.sub(rf'<codeclass="language-{codeclass}">.*</pre>', "", html1)
-        html2 = re.sub(rf'<codeclass="language-{codeclass}">.*</pre>', "", html2)
+    ignore_classes = [f"language-{lang}" for lang in enabled_codeformatters]
+    mdit = MarkdownIt()
+    for name in enabled_extensions:
+        plugin = mdformat.plugins.PARSER_EXTENSIONS[name]
+        plugin.update_mdit(mdit)
+        ignore_classes.extend(getattr(plugin, "ignore_classes", []))
```
hukkinj1, line 26: This is new plugin API and should be documented if we leave it in. What would this be used for though? There aren't many classes in HTML generated from Markdown.

I've been thinking to make a TOC generator plugin where I would need to ignore everything in the HTML that's in between certain types of HTML comments. A regex based ignore system could work there, or a flag that disables equality check altogether, but not sure if a class based system like this would.

hukkinj1 (changes requested):

mdformat/_util.py
```diff
@@ -1,21 +1,81 @@
...
+            self.tree.append(tag_item)
+        else:
+            children = self.current.setdefault("children", [])
+            children.append(tag_item)
+        self.current = tag_item
+
+    def handle_endtag(self, tag: str) -> None:
+        # walk up the tree to the tag's parent
+        while self.current is not None:
+            if self.current["tag"] == tag:
```
hukkinj1, line 69: Is it possible that this exits too early here, if say a `<ul>` tag is nested inside another `<ul>` tag?

mdformat/_util.py
chrisjsewell: Well yes it does produce valid JSON, which also happens to be an AST of sorts. So really either is apt.
Note that the other approach here is just to use https://www.crummy.com/software/BeautifulSoup/bs4/doc/ (that's the code I originally wrote), i.e. having an extra (albeit very well known) dependency versus using a bespoke parser that requires a little more validation testing. I'll leave you to decide which?

mdformat/renderer/_token_renderers.py
chrisjsewell: If we are addressing later it in #37 then yeh sure

mdformat/_util.py
chrisjsewell: yeh I guess it just depends on how rigorous you want to be. Happy to drop

mdformat/_util.py
chrisjsewell: well self.current can technically be `None` (although it will have been replaced in `handle_starttag` by this point) and so mypy complained about it

mdformat/renderer/_token_renderers.py
hukkinj1: I refer to line 111, not 110 :smile:

mdformat/_util.py
chrisjsewell: why in between HTML comments? what would the ToC HTML structure look like?
and anyway, if we drop `handle_comment` (above), then comments would be ignored anyway?

well you could have something like the amsmath plugin that generates:

```html
<section class="amsmath">
<eqn>
\begin{equation*}
a = 1
\end{equation*}
</eqn>
</section>
```

and so if you wanted to format the TeX in some way, you would specify `ignore_classes = ["amsmath"]`. You could even take it one step further and specify CSS selector like syntax, e.g. `section.amsmath` (this is available in beautifulsoup)

mdformat/renderer/_token_renderers.py
chrisjsewell: oh yeh oops

mdformat/_util.py
hukkinj1: Comments can't have any effect on rendered HTML so IMO we don't need to care about them here.

mdformat/renderer/_token_renderers.py
chrisjsewell: we can probably improve on this anyway, but will leave it in there for a start

mdformat/_util.py
hukkinj1: I believe JSON is a text format, so we'd need to do `json.dumps()` to the output before we can call it JSON.

I think this custom parser should be just fine if we can keep it reasonably compact.

chrisjsewell: yep fair

mdformat/_util.py
hukkinj1: Aah if its just mypy complaining, I'd help mypy with a `assert self.current is not None` instead of the `if` clause

mdformat/_util.py
chrisjsewell: No pretty sure not, because it first exits the inner </ul> then the outer `</ul>`.
I can add an extra test though.

mdformat/_util.py
chrisjsewell: I can add it to the `ParserExtensionInterface` though

hukkinj1: > what would the ToC HTML structure look like?

Something like https://github.com/thlorenz/doctoc#specifying-location-of-toc

So you leave these comments in the Markdown and the tool will generate a TOC in between them.

hukkinj1: Hmm, I'm thinking maybe just a boolean switch to turn off the Markdown equality check altogether might be the cleanest solution that would work for every possible use case, and with very little effort...?

chrisjsewell: yeh that feels like a bit of an edge case that we can deal with later

chrisjsewell:
applied changes

mdformat/_util.py
chrisjsewell: added test

mdformat/_util.py
```diff
@@ -1,21 +1,78 @@
...
+        plugin.update_mdit(mdit)
+        ignore_classes.extend(getattr(plugin, "ignore_classes", []))
+    html1 = HTML2AST().parse(mdit.render(md1), ignore_classes)
+    html2 = HTML2AST().parse(mdit.render(md2), ignore_classes)
+
     return html1 == html2
+
+
+class HTML2AST(HTMLParser):
+    """Parser HTML to JSON."""
```
hukkinj1, line 34: JSON --> AST rename

mdformat/_util.py
```diff
@@ -1,21 +1,78 @@
...
+            if self.current["tag"] == tag:
+                self.current = self.current.pop("parent")
+                break
+            self.current = self.current.pop("parent")
+
+    def handle_data(self, data: str) -> None:
+        # ignore empty data
+        if data.strip():
+            assert self.current is not None, f"'{data}'"
+            self.current["data"] = data.rstrip()
```
hukkinj1, line 78: Is there a reason why `strip` above and `rstrip` here? Maybe `strip` both?

mdformat/_util.py
chrisjsewell: BTW it feels like there could be some possible integration of the ToC plugin and https://github.com/executablebooks/markdown-it-py/blob/master/markdown_it/extensions/anchors/index.py

chrisjsewell: i.e. potentially moving the anchor plugin code out of markdown-it-py and into the same package as the toc generator plugin

hukkinj1: Ah so what I'm suggesting already now, before we add the class-based-skip-equality-check API, should we just instead give plugins a boolean flag they can set that completely disables the equality check. The ToC plugin for instance seems to need something more powerful than the class-based system, and I wouldn't want to end up supporting many ways to bypass the equality check.

This way:
- the API is dead simple
- we never have to extend it

The whole equality check is just a safeguard really, intended for the default "pure" formatter mode, that is not supposed to alter AST.

hukkinj1 (changes requested):

mdformat/_util.py
chrisjsewell: done

mdformat/_util.py
chrisjsewell: I've gone back to ignoring data outside tags (i.e. ignore `self.current = None`) so that it ignores any newline markers at the start/end of the file.
I've also changed to splitting into lines and removing all empty lines and trailing whitespace.
I think this is good enough to check for a high degree of equality.

mdformat/_util.py
chrisjsewell: I prefer to keep the equality check where possible. I think for most plugins the `ignore_classes` option should be sufficient.
This ToC plugin feels like a bit of special case, in that it is actually adding something new to the document, rather than just formatting what is already there

hukkinj1: Ok so I thought about this a bit. Current thoughts:

- I will want to add a `CHANGES_AST` (or similar name) switch to the plugin API in any case.
- The equality check is really just a safeguard for the CLI (currently not applied to Python API), not a core part of mdformat. Assuming no bugs, the feature does nothing. I want to avoid overengineering a set of plugin hooks to alter that.
- The formatter in theory should never alter AST/rendered markdown. I feel like any plugin that does so, is equally a special case.

So I'm thinking to reject the class-based plugin hook. I know this could be a bit frustrating, sorry for that. And if so, I encourage to plan new features in an issue before implementation.

chrisjsewell: could the equality check be toggled to only check the equality of the HTML AST structure, rather than its full content, i.e. `def handle_data` would not store anything in this mode.
This would still check that the formatting does not change the overall tag structure of the HTML (which a plugin should not generally do), but would not worry about its content (which mdformat core and plugins literally all change).

e.g. it would just check that both docs have the structure

```html
<div a="1">
  <other b="2">
  </other>
</div>
```

rather than check that they are exactly:

```html
<div a="1">
  a b
  c d
  <other b="2">
  x y z
  </other>
</div>
```

chrisjsewell: > The formatter in theory should never alter AST/rendered markdown

and yeh just to emphasise; in some sense the formatter *always* alters the HTML AST, if you are taking into account the content of each tag, e.g. stripping whitespace in this context is altering the AST.

what the formatter should not do (in theory) is alter the HTML AST structure

hukkinj1: > could the equality check be toggled to only check the equality of the HTML AST structure, rather than its full content

This would also require an addition to the API, which I'd rather not do. I've thought about possibly even removing the equality check once mdformat proves to be stable and bug free (since it makes the tool render 3 docs instead of just 1). I'd prefer the API to not be littered with stuff related.

> This would still check that the formatting does not change the overall tag structure of the HTML (which a plugin should not generally do), but would not worry about its content (which mdformat core and plugins literally all change).

> stripping whitespace in this context is altering the AST.

I mean CommonMark AST, which AFAIK should completely ignore leading or trailing whitespace. The HTML rendering is just a tool used to check that.
Besides leading/trailing whitespace the changes mdformat does are mostly just adding escape characters and converting characters to their numeric representation (which AFAIK should not change HTML in any way).

> what the formatter should not do (in theory) is alter the HTML AST structure

It's more strict than this. All ASTs aside, what I really want is the rendered HTML to look precisely the same before and after running mdformat (core). That means that the changes mentioned above are fine (when applied to paragraphs), but most other changes to content (e.g. any changes to code blocks) are not. Plugins are free to do what they want of course.

chrisjsewell: > I've thought about possibly even removing the equality check

I don't know, I quite like it. You probably already know, but black has a similar equivalence check
https://github.com/psf/black/blob/172c0a78facd8f938629f216c95507bbfeddfe5e/src/black/__init__.py#L944

> I mean CommonMark AST, which AFAIK should completely ignore leading or trailing whitespace.

I'm not sure if this is correct. I mean there's not really a CommonMark AST anyway, just a specification of what HTML it should produce.

It would just be nice if you didn't have to completely throw away this check, due to the use of one plugin.

hukkinj1: > I'm not sure if this is correct. I mean there's not really a CommonMark AST anyway, just a specification of what HTML it should produce.

I'm pretty sure every language can be parsed into a syntax tree, even if there's not an established format how it should look like. According to CommonMark spec `The paragraph’s raw content is formed by concatenating the lines and removing initial and final whitespace.`, so a parser will definitely remove leading/trailing whitespace and it won't be in this syntax tree.

> It would just be nice if you didn't have to completely throw away this check, due to the use of one plugin.

With the `CHANGES_AST` flag, you don't have throw it away, unless the plugin makes changes to how the rendered Markdown looks like. That seems like simple and reasonable spec to me.

I think at this point we should just agree to disagree? :smile: :peace_symbol:

hukkinj1:
I will close this PR as I don't want the class based equality check ignore. Without it I feel like the HTML parser might also not be necessary. The tests here I definitely want though. Many of the test changes, and the warning log added here seem to already be in `master`.

I've taken the commits, applied changes from `master` and cut out the changes that seem unnecessary. The result is here https://github.com/executablebooks/mdformat/pull/66

The tests found one bug in `master`, thank you very much @chrisjsewell !

Issue #33, opened 2020-09-25
Consecutive numbered lists


chrisjsewell:
I can see the thinking for using `1.` for all numbered list items; with a renderer handling proper numbering.
However, one of the benefits of Markdown is that it is easily readable without the need for rendering.
This is diminished with this sort of (non-)numbering, and I bet most people would opt for consecutive numbering, if given the choice.

I would say there should at least be the option (if not in "core" then *via* a plugin) to achieve this.

hukkinj1:
For me some of the objectives/goals in mdformat development have been (heavily inspired by e.g black) (maybe I should write something in the contribution guide about this):
- mdformat attempts to minimize diffs
- mdformat is opinionated so you don't have to be
- no configuration unless a very good reason

This (non-)numbering style was done on purpose to minimize diffs. Lets say I'm listing the alphabets, using a style where list items are correctly numbered:

```markdown
1. b
2. c
3. d
```

I then notice I forgot the character "a" and add it as first item in the list. As a result, the numbering of every subsequent item in the list will change, meaning that the diff will touch every line in the list. The non-numbering style solves this issue: only the added line will show up in the diff.

In this case the style comes at the cost of some readability. The nice thing about Markdown is, as you mentioned, if we really need it to be readable, we can always render the HTML.

So with all this in mind:
- the default non-numbering style won't change
- I'm a bit reluctant still :smiley: , but this could be one of the exceptions where adding configuration to "core" might be ok
- a plugin that does numbering should be quite straightforward to implement with the current plugin API already

chrisjsewell:
thanks for the explanation, yep I didn't consider the "minimize diffs" argument, which is an interesting point.
Those goals should definitely be documented 👍

hukkinj1:
I considered this a bit. I'd be willing to merge a CLI flag that changes the style to consecutive numbering. The default style will remain the same though.

chrisjsewell:
should be easy to add once #35 is merged

PR #38, opened 2020-09-27
✨ NEW: Keep link references

chrisjsewell:
Utilise the new `store_labels` option I introduced in mardown-it-py, to keep link/image reference labels, instead of their resolved href. Then output the (sorted) references at the end of the document.

Notes:
- I've left a "hook" to turn this feature on/off: `markdown_it.options["mdformat"] = {"keep_references": True}`, this could be turned into a CLI option, if desired.
- References are always normalized using `str.upper()` in markdown-it, but, as a style choice, I convert them to `str.lower()`
- I've added a `finalize` key-word argument to `MDRenderer.render`. This is for use by plugins (like mdformat-tables) that need to run "nested rendering", during which they do not want these references output.

### DIFF
--- a/mdformat/renderer/__init__.py
+++ b/mdformat/renderer/__init__.py
@@ -111,14 +111,14 @@ class MDRenderer:
 
             if env.get("used_refs", None):
                 rendered_content += "\n\n"
-                rendered_content += self.write_references(env)
+                rendered_content += self._write_references(env)
 
             rendered_content += "\n"
 
         return rendered_content
 
     @staticmethod
-    def write_references(env: dict) -> str:
+    def _write_references(env: dict) -> str:
         text = ""
         for key in sorted(env.get("used_refs", [])):
             ref = env["references"][key]
--- a/mdformat/renderer/__init__.py
+++ b/mdformat/renderer/__init__.py
@@ -109,9 +109,7 @@ class MDRenderer:
             rendered_content = removesuffix(rendered_content, MARKERS.BLOCK_SEPARATOR)
             rendered_content = rendered_content.replace(MARKERS.BLOCK_SEPARATOR, "\n\n")
 
-            if options.get("mdformat", {}).get("keep_references", False) and env.get(
-                "references", None
-            ):
+            if env.get("used_refs", None):
                 rendered_content += "\n\n"
                 rendered_content += self.write_references(env)
 
@@ -122,7 +120,7 @@ class MDRenderer:
     @staticmethod
     def write_references(env: dict) -> str:
         text = ""
-        for key in sorted(env.get("references", {}).keys()):
+        for key in sorted(env.get("used_refs", [])):
             ref = env["references"][key]
             item = f"[{key.lower()}]: {ref['href']}"
             if ref["title"]:
--- a/mdformat/renderer/__init__.py
+++ b/mdformat/renderer/__init__.py
@@ -126,6 +126,6 @@ class MDRenderer:
             ref = env["references"][key]
             item = f"[{key.lower()}]: {ref['href']}"
             if ref["title"]:
-                item += f'"{ref["title"]}"'
+                item += f' "{ref["title"]}"'
             text += item + "\n"
         return text.rstrip()
--- a/mdformat/renderer/__init__.py
+++ b/mdformat/renderer/__init__.py
@@ -27,6 +27,7 @@ class MDRenderer:
         options: dict,
         env: dict,
         *,
+        finalize: bool = True,
         _recursion_level: int = 0,
     ) -> str:
         """Takes token stream and generates Markdown.
@@ -35,6 +36,7 @@ class MDRenderer:
             tokens: A list of block tokens to render
             options: Params of parser instance
             env: Additional data from parsed input
+            finalize: replace markers and write references
         """
         assert _recursion_level in {
             0,
@@ -66,7 +68,11 @@ class MDRenderer:
             if token.type == "inline":
                 # inline tokens require nested rendering
                 result = self.render(
-                    token.children, options, env, _recursion_level=_recursion_level + 1
+                    token.children,
+                    options,
+                    env,
+                    finalize=finalize,
+                    _recursion_level=_recursion_level + 1,
                 )
             else:
                 # otherwise use a built-in renderer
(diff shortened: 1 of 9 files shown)


chrisjsewell:
FYI, whats the reasoning for always wrapping link URLs in `<>`? https://github.com/executablebooks/mdformat/blob/200aaea3d408582c65fdeffd60f46822e8436495/mdformat/renderer/_token_renderers.py#L37

hukkinj1:
> Also, whats the reasoning for always wrapping link URIs in <>? ... If I recall correctly, this signifies an external link

I don't think CommonMark spec makes a distinction between external or internal links. With the `<>` brackets the spec allows a wider set of characters and less character escaping than without. So the formatting logic is a lot simpler with the brackets, and all links look the same, those are the reasons for the wrapping.

hukkinj1:
> I've left a "hook" to turn this feature on/off: markdown_it.options["mdformat"] = {"keep_references": True}, this could be turned into a CLI option, if desired.

As a bit of a minimalist, I would prefer to remove this hook. Let's just change the default style to "keep ref links as ref links and inline links as inline links".

> References are always normalized using str.upper() in markdown-it, but, as a style choice, I convert them to str.lower()

Very good, i like this.

hukkinj1:
I would prefer mdformat to delete unused refs (similar to how it deletes duplicate refs). Might want to make a test case that checks that none of these "extra newlines before references" are left at the end of file if there's zero refs.

mdformat/renderer/__init__.py
```diff
@@ -99,8 +105,27 @@ def render(
...
+            ):
+                rendered_content += "\n\n"
+                rendered_content += self.write_references(env)
+
             rendered_content += "\n"
+
         return rendered_content
+
+    @staticmethod
+    def write_references(env: dict) -> str:
```
hukkinj1, line 123: Does this need to be exported? I'd prefer it to be private method if not.

tests/test_renderer_style.py
```diff
@@ -1,45 +1,18 @@
-from markdown_it import MarkdownIt
+from pathlib import Path
+
+from markdown_it.utils import read_fixture_file
 import pytest
 
-from mdformat.renderer import MDRenderer
+from mdformat import text as render
```
hukkinj1, line 6: What's the reason for this change? I would perhaps prefer to leave this as a test for the `renderer` package. There is another one for the public API functions.

tests/test_renderer_style.py
```diff
@@ -1,45 +1,18 @@
...
+    "line,title,text,expected", STYLE_CASES, ids=[f[1] for f in STYLE_CASES]
+)
+def test_renderer_style(line, title, text, expected):
     """Test Markdown renderer renders expected style."""
-    md_original = entry["input_md"]
-    md_new = MarkdownIt(renderer_cls=MDRenderer).render(md_original)
-    expected_md = entry["output_md"]
-    assert md_new == expected_md
+    md_new = render(text)
+    print(md_new)
```
hukkinj1, line 17: Should we remove the print?

chrisjsewell:
> I don't think CommonMark spec makes a distinction between external or internal links. With the <> brackets the spec allows a wider set of characters and less character escaping than without. So the formatting logic is a lot simpler with the brackets, and all links look the same, those are the reasons for the wrapping.

As with #33 I suggest these type of considerations are documented, since I'm sure I won't be the only one question this.

I guess then that this should be the same for image uri?

tests/test_renderer_style.py
chrisjsewell: I'm going to leave it in, but only when the assertion fails, since it's a very handy practice for adding/modifying tests, to add a new test input, run pytest, then copy/paste the printed output to the expected section.

tests/test_renderer_style.py
chrisjsewell: I've changed it pack to using the renderer, but note you have to add:

```python
md_new.options["store_labels"] = True
```

for the references to be recovered (see https://github.com/executablebooks/markdown-it-py/commit/d9b6de9c8ff3bb0dc2b78d6eda8e09b3b88fa7b4)

mdformat/renderer/__init__.py
chrisjsewell: Why not allow it to be overridden in a subclass?
In fact, technically I'd say it's a bit of a shortcoming of this renderer; that you can't override any of the builtin token render methods in a subclass, as opposed to https://github.com/executablebooks/markdown-it-py/blob/master/markdown_it/renderer.py where the render method for every token can be overridden.
There's not an immediate use case for that though, and you can now achieve it via plugins, so I'm not suggesting changing it, just noting 😉

chrisjsewell:
> As a bit of a minimalist, I would prefer to remove this hook. Let's just change the default style to "keep ref links as ref links and inline links as inline links".

done

> I would prefer mdformat to delete unused refs (similar to how it deletes duplicate refs). 

done

> Might want to make a test case that checks that none of these "extra newlines before references" are left at the end of file if there's zero refs.

new lines and references are only added if there are any referenced labels, so this is not an issue

tests/test_renderer_style.py
hukkinj1: Ah alright. What happens if `store_labels = False` now? The ref links just get formatted as inline (which would seem fine). If it's something more unexpected, we might want to just fail in `MDRenderer.__init__` with a ValueError if store_labels isnt enabled.

chrisjsewell: yep they just get formatted as inline 👍

hukkinj1:
> As with #33 I suggest these type of considerations are documented, since I'm sure I won't be the only one question this.

:+1: made an issue for this

> I guess then that this should be the same for image URIs?

I would imagine so as well. I'm pretty sure there's some test that breaks if the angle brackets are removed from normal links. We should probably make similar tests as the failing ones, but for images, and see what happens. Making an issue.

mdformat/renderer/__init__.py
hukkinj1: So by default I always want to make everything private to avoid a maintenance nightmare. Any changes, refactoring etc. that change public API is a breaking change and forces me to make a major version bump, and is a maintenance hassle for a potential user.

> In fact, technically I'd say it's a bit of a shortcoming of this renderer;

It is a shortcoming for someone willing to subclass the class. It is also a major advantage for me trying to maintain the project. So far someone subclassing the renderer class is hypothetical, but me maintaining it is reality.

I didn't really intend to make even `MDRenderer.render` public API, it just happened to be useful for the plugin interface. I mostly care about having the formatter API public.

hukkinj1: I'm not fully opposed to making any API public, I just think there should usually be a good use-case and hopefully also at least one user willing to use it.

chrisjsewell: sure, its not a hill I want to die on lol; changed in 895d3106338c543d4267f5976af7978a3fdecaa6

hukkinj1: :smile: nice thanks!

hukkinj1:
Very nice, thanks @chrisjsewell !! Merged

PR #41, opened 2020-09-30
✨ Add consecutive numbering option

### LINKED ISSUE #33: Consecutive numbered lists

chrisjsewell:
Allows for application of consecutive numbering to ordered lists

Note this is currently rebased onto #35 

closes #33

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ from markdown_it.token import Token
 from mdformat.renderer import MDRenderer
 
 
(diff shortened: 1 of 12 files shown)


hukkinj1:
Awesome, thanks @chrisjsewell ! Merged

PR #46, opened 2020-10-01
Make code block stripping regex make sense

hukkinj1:


mdformat/_util.py
```diff
@@ -27,9 +28,9 @@ def is_md_equal(
         mdformat.plugins.PARSER_EXTENSIONS[extension].update_mdit(mdit)
     for key, text in [("md1", md1), ("md2", md2)]:
         html = mdit.render(text)
-        html = re.sub(r"\s+", "", html)
+        html = re.sub(r"\s+", " ", html)
```
chrisjsewell, line 31: This is a pretty big change, and will definitely break tables, where you can have `|a|` go to `| a |`.
What if you stripped whitespace from the end of a line?

hukkinj1: Not sure I understand. How does `|a|` go to `| a |`? The regex should only match one or more whitespace chars. If there are none, it shouldn't create them out of nothing.

chrisjsewell: Well firstly, if you have and input (`.` is a space)

```
a.
```

the output would be

```
a
```

so now these would surely be non-equivalent, because that space would remain?

for tables, you have:

```
|a|
|-|
|b|
```

got to:

```
| a |
| - |
| b |
```

so now these are non-equivalent

chrisjsewell: I'm not definitely sure if I'm right about this, but feel it definitely needs testing, and is not a trivial change

hukkinj1: Thanks for pointing this out. I reverted the whitespace change. What do you think of the PR now? 

I might get back to the whitespace thing later with improved tests. Preserving the one whitespace makes the check slightly stricter for sure, as it differentiates between a whitespace sequence and no whitespace sequence. I think we might want that strictness though. Not sure if the table example is a problem, as the table will be rendered to an HTML table at this point.

chrisjsewell: yeh PR looks fine now. Just wanted to make sure we were rushing into anything breaking 😄

chrisjsewell (approved):

Issue #39, opened 2020-09-27
Document some decisions


hukkinj1:
See https://github.com/executablebooks/mdformat/pull/38#issuecomment-699656425 and https://github.com/executablebooks/mdformat/issues/33#issuecomment-699525575

PR #73, opened 2020-10-20
Styleguide

hukkinj1:
Closes https://github.com/executablebooks/mdformat/issues/39

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -83,7 +83,7 @@ jobs:
         pip install pre-commit
 
     - name: run hook against repository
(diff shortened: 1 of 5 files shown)


chrisjsewell:
Looking good 👍 
Maybe mention the reason for surrounding links in `<>`, e.g. `[my link](<https://example.com>)`

hukkinj1:
> Maybe mention the reason for surrounding links in <>, e.g. [my link](<https://example.com>)

You're reading my mind.

Although, I couldn't come up with good enough reasons so opted for making this instead https://github.com/executablebooks/mdformat/pull/74

:smile:

chrisjsewell:
> I couldn't come up with good enough reasons so opted for making this instead

Ah even better 😄

PR #129, opened 2021-02-15
renderer: Add a "translator" option

jpmckinney:
Newlines in paragraph text are replaced with spaces, and text is stripped of left and right whitespace, to have the same behavior as in Sphinx: https://github.com/sphinx-doc/sphinx/blob/v3.4.0/sphinx/util/nodes.py#L276

The `translator` option would usually be the return value of `gettext.translation`: https://docs.python.org/3/library/gettext.html#gettext.translation


jpmckinney:
I'm not familiar with the tests. If you can point me to a test that is broadly similar, I can add a new test.

hukkinj1:
Hey, please see [this comment](https://github.com/executablebooks/MyST-Parser/issues/311#issuecomment-779442490) first. There could be a better way.

PR #132, opened 2021-02-16
Tree structured rendering. New parser extension plugin API

hukkinj1:
TODO:

- [x] Redesign parser extension plugin API
  - [x] enable tests (currently marked `xfail`)
  - [x] update docs
- [x] Tidy up
- [x] Consider an alternative solution once more?
  - [x] Rebase https://github.com/executablebooks/mdformat/pull/85 on this branch and ensure that this actually helps there
- [x] Write changelog
  - `mdformat.renderer.MARKERS` removed
  - parser extension plugin API changed
    - Removed `mdformat.plugins.ParserExtensionInterface.render_token`
    - Added `mdformat.plugins.ParserExtensionInterface.RENDERER_FUNCS`
  - `start` and `stop` kwargs removed from `mdformat.renderer.MDRenderer.render`, use `render_tree` to render partial Markdown
  - `mdformat.renderer.TreeNode` added
  - `mdformat.renderer.MDRenderer.render_tree` added
- [ ] wait for markdown-it-py > 0.6.2 to be released on PyPI
- [ ] change markdown-it-py git dependency to a PyPI dependency in pyproject.toml. `poetry update` after


chrisjsewell:
Just out of interest, why this change?

hukkinj1:
My thinking is this might be the least hacky solution to the problems I've ran into working on https://github.com/executablebooks/mdformat/pull/85 (or in general, modes where original word wrap is not preserved).

https://github.com/executablebooks/mdformat/pull/85 lists some of the problem points. In brief, not all whitespace is equal in Markdown, so when altering word wrap I need some context (e.g. is the whitespace in an HTML block, part of an inline code block etc.).

The current paragraph formatter that should do word wrap changes (`def paragraph_close` function) only receives a paragraph `str` and is not aware what parts of the `str` belong in what Markdown structures. This limitation has been blocking my work.

A very unfortunate side effect of a change like this is that I think I'd have to break the plugin API...

This is still sort of a "conceptualization phase" so if you have any thoughts, ideas, or alternative ways, that could help out, would appreciate it :smiley_cat:

chrisjsewell:
cheers will do

hukkinj1:
@chrisjsewell I still haven't been able to think of a less disruptive way to tackle the issues I had, so I'm thinking to merge this.

This will break parser extension plugin API just as I feared earlier. I already sketched out a PR to adapt mdformat-tables to the changes: https://github.com/executablebooks/mdformat-tables/pull/7

If you have time and interest to do a review, would be great! Otherwise I think I'll merge in a day or two.

chrisjsewell:
thanks for the heads-up @hukkinj1, I'll try to find some time to review, but don't worry if not.

> This will break parser extension plugin API just as I feared earlier.

can you clarify what you mean by this; does it reduce the functionality available to plugins, or is it just that they have to be re-written?

hukkinj1:
> can you clarify what you mean by this; does it reduce the functionality available to plugins, or is it just that they have to be re-written?

Just have to be rewritten. There shouldn't be any reduced functionality. The plugins just now have to deal with the `SyntaxTreeNode` tree structure instead of the linear `List[Token]` stream.

mdformat/renderer/__init__.py
```diff
@@ -142,5 +89,125 @@ def _write_references(env: dict) -> str:
             if title:
                 title = title.replace('"', '\\"')
                 item += f' "{title}"'
-            text += item + "\n"
-        return text.rstrip()
+            ref_list.append(item)
+        return "\n".join(ref_list)
+
+
+class SyntaxTreeNode:
```
chrisjsewell, line 96: so this is just a reworking of `NestTokens` and `nest_tokens` (see https://github.com/executablebooks/markdown-it-py/blob/2afb7cc5a824cf8340e405704cc996208ae33d0e/markdown_it/token.py#L114). I think it should just go into markdown-it-py and deprecate those

mdformat/renderer/__init__.py
```diff
@@ -142,5 +89,125 @@ def _write_references(env: dict) -> str:
...
+          tokens nested in between
+    """
+
+    def __init__(self) -> None:
+        """Initialize a root node with no children.
+
+        You probably need `SyntaxTreeNode.from_tokens` instead.
+        """
+        # Root and containers don't have self.token
+        self.token: Any = None  # Optional[Token]
```
chrisjsewell, line 115: why are these set to `Any`?

mdformat/renderer/__init__.py
hukkinj1: Yeah precisely so. Why I didn't use `NestedTokens` is explained in this closed draft PR https://github.com/executablebooks/markdown-it-py/pull/117

> I think it should just go into markdown-it-py and deprecate those

That would be great, but it will mean some more work for you :smile: I understood `NestedTokens` is a central piece in MyST parser. But I'm very much willing to move this class to `markdown_it` if you think its a good idea. I think the `.render` method should probably be removed then, and I'll subclass and add it back here in `mdformat`. That method is quite `mdformat` and rendering specific...

mdformat/renderer/__init__.py
hukkinj1: This is because the class is very dynamic, not a very good match for static type checking.

The type (`None` or `Token`) depends on whether the node represents a root node, an unnested `Token` or a nested "_open" and "_close" `Token` pair. This information is dynamic though, and the class is always the same in the three different cases. If I type annotate `Optional[Token]` I will have to add a thousand `assert`s to any code that uses this class to prove the type checker what the actual type is.

mdformat/renderer/__init__.py
chrisjsewell: > I understood `NestedTokens` is a central piece in MyST parser

yes but we can just deprecate it, leaving it in the code base for now, and I can migrate in due course

mdformat/renderer/__init__.py
chrisjsewell: I think it should work similar to what `NestedToken` does now, in that you add some "first class" attribute/methods mimicking the `Token` and then these are either taken from self.token or self.opening, e.g something like.

```python
    @property
    def map(self):
		if self.token:
            return self.token.map
        elif self.opening:
            return self.opening.map
        else:
            return None
```

then the user code doesn't have to do any of this checking and can just treat it similar to a `Token`.

chrisjsewell: an `is_nested` property would also be helpful:

```python
@property
def is_nested(self) -> bool:
    return self.opening is not None
```

hukkinj1: You'll have the same problem with dynamic types with the typing of `SyntaxTreeNode.map` then though, i.e. will have to keep it's return type as `Any`.

I'm a bit in two minds when it comes to adding many convenience methods like that. Keeping the abstraction leaner (not hiding the fact that it's just an alternative structurer for a `List[Token]`) has the advantage that it will work the same way with *__any__* `Token` type, whether it's defined by `MarkdownIt` itself or any arbitrary plugin/extension.

chrisjsewell: > You'll have the same problem with dynamic types with the typing of SyntaxTreeNode.map then though, i.e. will have to keep it's return type as Any.

why?

> that it will work the same way with any Token type

what do you mean by any here? in markdown-it all tokens are the same, i.e. you always just use `Token`, thats really one of the core design philosophies of markdown-it

hukkinj1: Regarding `is_nested`, I agree that could be useful, and is very general purpose. Maybe also `is_root`? That can be checked with `node.parent is None` though... But a set of methods that makes it clear that there's three basic "types" (root, nested, unnested) makes sense.

hukkinj1: I'm also thinking to have `SyntaxTreeNode.to_tokens()` as a reverse for `SyntaxTreeNode.from_tokens()`. Should be very easy to implement.

chrisjsewell: on a similar point users should not be able to modify self.token, self.opening and self.closing, i.e. they should really be private variables. Because you essentially should never allow both self.token and self.opening to be set at the same time; its either nested or its not nested

chrisjsewell: > I'm also thinking to have SyntaxTreeNode.to_tokens() as a reverse for SyntaxTreeNode.from_tokens(). Should be very easy to implement.

indeed

hukkinj1: > why? why can't it be typed as Optional[List[int]]

So the same thing as with all the attributes typed as `Any` in this class currently. It __can__ be done of course, but the type (`None` or `List[int]`) depends on dynamic information that the type checker is unaware of. So when writing code, every time you use `SyntaxTreeNode.map` somewhere, say you checked `is_nested` value so you are 100% sure the type of `SyntaxTreeNode.map` is `List[int]`, you still have to prove it to the type checker with an `assert is not None`, every single time.

Since the class is dynamic, a static type checker is more a nuisance than helpful.

chrisjsewell: > you still have to prove it to the type checker with an assert is not None, every single time.

you already have to do this because `Token.map` is already of type `Optional[List[int]]`: https://github.com/executablebooks/markdown-it-py/blob/2afb7cc5a824cf8340e405704cc996208ae33d0e/markdown_it/token.py#L20

hukkinj1: > on a similar point users should not be able to modify self.token, self.opening and self.closing, i.e. they should really be private variables. Because you essentially should never allow both self.token and self.opening to be set at the same time; its either nested or its not nested (or root)

:+1: agree that users should never modify. But not sure we should  deny read access, as then some information in some arbitrary `Token` that we didn't build custom handling for might be unavailable.

We could keep it public and document that token is read-only, and trust the users. Or make it a private attribute with a public `@property` with no setter.

hukkinj1: > you already have to do this because Token.map is already of type Optional[List[int]]

:+1: True, my personal opinion is it would be better to type that as `Any` and leave a comment what the real (dynamic) typings are.

chrisjsewell: I would be tempted to do something like this

```python
class SyntaxTreeNode:
    def __init__(self, token=None, opening=None, closing=None) -> None:
        assert not (token is not None and opening is not None)
        if token is None and opening is None:
             # it is root
             self.token = Token("root")
        else:
            self.token = token
            self.opening = opening
            self.closing = closing
        ...

    @propery
    def _attribute_token(self) -> Token:
        if self.token is not None:
            return self.token
        assert self.opening is not None
        return self.opening

    @propery
    def map(self) -> Optional[List[int]]:
      return self._attribute_token.map

    @propery
    def info(self):
      return self._attribute_token.info

    ...

```

chrisjsewell: this avoids any dynamic types, since they will always be exactly the same as for `Token`

hukkinj1: > But not sure we should deny read access, as then some information in some arbitrary Token that we didn't build custom handling for might be unavailable.

To add to this: imagine a case where the node represents a nested token pair, and some data is only available in the closing token, not opening token. It gets real messy to provide access unless we expose read access to both opening and closing token.

A leaner abstraction might work better here.

chrisjsewell: > imagine a case where the node represents a nested token pair, and some data is only available in the closing token, not opening token. 

this is not the case for any existing Token pair. We can provide access to the closing token just in case, but thus far it is not used anywhere in this code or anywhere in myst-parser

hukkinj1: I kind of like the example code you made above. Besides the (admittedly theoretical :smile: ) closing token issue.

Btw, what's your take on differentiating from JS typings slightly to make things more static? Example: Make `Token.children` always a `List[Token]` instead of an optional. Means the default value will need to be changed from `None` to `[]`.

Btw Github issue is a pretty bad place for fast paced discussion like this. Need an old-school IRC channel, lol

chrisjsewell: > Need an old-school IRC channel, lol

lol well we do have an EPB Slack channel you are welcome to join

> Btw, what's your take on differentiating from JS typings slightly to make things more static? 

I'm not totally against it, but I am always at bit hesitant in case it has side-effects, and then I keep putting off reviewing/approving it like https://github.com/executablebooks/markdown-it-py/pull/102 lol

hukkinj1: > lol well we do have an EPB Slack channel you are welcome to join

Could possibly be handy in the future. I guess I need an invite? My address is first name dot family name and the domain is google's free mail service.

> I'm not totally against it, but I am always at bit hesitant in case it has side-effects, and then I keep putting off reviewing/approving it like

Understood. And don't worry about it, not the world's most important PR :smile:

hukkinj1: I'm marking this thread resolved now that `SyntaxTreeNode` is merged to `markdown-it-py` https://github.com/executablebooks/markdown-it-py/pull/129

PR #142, opened 2021-03-09
Adding reference to mdformat-frontmatter as an available plugin

butler54:
I recently created a plugin to allow mdformat to handle yaml headers used by some tools. This PR is to add that plugin to the documentation of mdformat.

Signed-off-by: Chris Butler <chris@thebutlers.me>

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,4 +260,9 @@ if you wish to implement a new parser extension plugin.
     <td><code>toc</code></td>
     <td>Adds the capability to auto-generate a table of contents</td>
   </tr>
+  <tr>
+    <td><a href="https://github.com/butler54/mdformat-toc">mdformat-frontmatter</a></td>
+    <td><code>frontmatter</code></td>
+    <td>Adds the capability parse yaml headers to markdown documents.</td>
+  </tr>
 </table>
--- a/README.md
+++ b/README.md
@@ -263,6 +263,6 @@ if you wish to implement a new parser extension plugin.
   <tr>
     <td><a href="https://github.com/butler54/mdformat-frontmatter">mdformat-frontmatter</a></td>
     <td><code>frontmatter</code></td>
-    <td>Adds the capability parse yaml headers to markdown documents.</td>
+    <td>Adds support for front matter, and formats YAML front matter</td>
   </tr>
 </table>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -260,4 +260,9 @@ if you wish to implement a new parser extension plugin.
     <td><code>toc</code></td>
     <td>Adds the capability to auto-generate a table of contents</td>
   </tr>
+  <tr>
+    <td><a href="https://github.com/butler54/mdformat-frontmatter">mdformat-frontmatter</a></td>
+    <td><code>frontmatter</code></td>
+    <td>Adds the capability parse yaml headers to markdown documents.</td>
```
hukkinj1, line 266: Suggestion:
```suggestion
    <td>Adds support for front matter, and formats YAML front matter</td>
```

hukkinj1 (changes requested):
Hey there!

I had a look at the plugin, looking good!

I'd like to suggest making the plugin not error given invalid YAML. Reasoning: I think generally pretty much any string should be valid Markdown, interpretation just changes depending on content. A formatter should generally not error on content that is valid Markdown, it should simply apply its formatting rules. Also the [front matter parser plugin](https://www.npmjs.com/package/markdown-it-front-matter) docs explicitly state that "YAML is not required".

Instead I suggest just passing through the non-YAML front matter text and logging a warning:
```python
import mdformat.renderer
mdformat.renderer.LOGGER.warning("Invalid YAML in a front matter block")
```
This log will printed to stderr if using the CLI.

butler54:
> Hey there!
> 
> I had a look at the plugin, looking good!
> 
> I'd like to suggest making the plugin not error given invalid YAML. Reasoning: I think generally pretty much any string should be valid Markdown, interpretation just changes depending on content. A formatter should generally not error on content that is valid Markdown, it should simply apply its formatting rules. Also the [front matter parser plugin](https://www.npmjs.com/package/markdown-it-front-matter) docs explicitly state that "YAML is not required".
> 
> Instead I suggest just passing through the non-YAML front matter text and logging a warning:
> 
> ```python
> import mdformat.renderer
> mdformat.renderer.LOGGER.warning("Invalid YAML in a front matter block")
> ```
> 
> This log will printed to stderr if using the CLI.

Hey there, 
Sorry for missing the comment. I'll have a look and update you when it's done with respect to the plugin.

butler54:
I updated the plugin based on your recommendations. It should be okay now.

Issue #72, opened 2020-10-20
Host documentation on readthedocs.io


hukkinj1:
- [ ] Create docs with either
      - MkDocs
      - Sphinx + CommonMark or MyST

- [ ]  Build the docs in Github Actions.

- [ ]  Publish on readthedocs.io

chrisjsewell:
Ideally Sphinx + MyST + RTD, the same as the rest of EBP 😬

You don't actually need to run a GH action, just turn on PR runs on RTD:
https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html

hukkinj1:
Thanks for the input, I really don't have much experience in any of this.

Sphinx + MyST + RTD sounds great. Or if we want to keep using `pre-commit try-repo ...` to run mdformat against its own docs, then maybe start with Sphinx + CommonMark and move to MyST, when we can come up with the necessary plugins for mdformat to handle MyST? I kinda like the `pre-commit try-repo` in that it shows we trust the tool enough to format our own docs with it, and also makes mdformat docs showcase what the output looks like.

> You don't actually need to run a GH action, just turn on PR runs on RTD:

Nice! Maybe it's possible to use a GH action to "validate" the docs though? Fail CI if docs fail building, so we know not to merge a PR unless the docs are fixed? Or does Sphinx build ever fail? I'm still a noob here :smile:

chrisjsewell:
> Sphinx + CommonMark and move to MyST

Oh MyST is already commonmark compliant. Its just add extra (optional) syntax plugins to the parser. If you don't use them, then it works exactly the same as CommonMark (we are in fact in talks with RTD to deprectate ReCommonMark in favour of MyST)

> Fail CI if docs fail building

Yep this already happens with the RTD builds (you can see the check e.g. in https://github.com/executablebooks/markdown-it-py/pull/58), you just need to add to the repo https://github.com/executablebooks/markdown-it-py/blob/master/.readthedocs.yml

chrisjsewell:
I can try to get round to making a PR for it at som point 😄

hukkinj1:
> Its just add extra (optional) syntax plugins to the parser.

Yeah this is what I was thinking could be problematic. But I think you are right, we don't really have to use the extra syntax and if we want to in some files, then skip mdformatting that file etc.

> Yep this already happens with the RTD builds

Nice!

> I can try to get round to making a PR for it at some point smile

This would be amazing!

PR #149, opened 2021-03-22
DOCS: Set up sphinx docs

hukkinj1:
Closes https://github.com/executablebooks/mdformat/issues/72

TODO:
- [x] Add actual content and split it into files based on structure. Currently I just copy-pasted the README example content.
- [ ] Make sure CI fails if sphinx build files
- [x] Make CI publish the docs
- [x] Move STYLE.md, CONTRIBUTING.md and maybe CHANGELOG.md to docs.
- [x] Simplify README.md after most of its content is in `docs/`
- [x] Current style is whatever the quickstart script had as default. Maybe change it?
- [ ] ~~Auto generated API docs?~~ maybe later
- [ ] Make readthedocs build `master` branch, not `docs` branch
- [x] Update [tool.poetry.urls] in pyproject.toml
- [x] Update `pre-commit try-repo` commands
- [ ] Add link to readthedocs link to GitHub "About"


hukkinj1:
@chrisjsewell Do you know if there's a way to make a ToC tree / navigation bar while having `myst_commonmark_only = True` set?

hukkinj1:
Drafting a possible docs structure here

- intro/summary/main page (reuse README.md here?)
  - overview
  - installing
- basic usage
  - cli
  - api
  - pre-commit
- plugin usage
- style guide
- contributing
  - code formatter plugins
  - parser extension plugins
- changelog

chrisjsewell:
> Do you know if there's a way to make a ToC tree / navigation bar while having `myst_commonmark_only = True` set?

Not currently, although actually jupyter-book does in that it uses a "top-down approach" for specifying the sitemap *via* a YAML file: https://jupyterbook.org/customize/toc.html 
I literally just discussed splitting this off into a separate module: https://github.com/executablebooks/meta/discussions/285#discussioncomment-518085

> recommonmark seems to be able to derive the toc from a CommonMark list

note it is eventually intended for recommonmark to be deprecated by myst-parser

hukkinj1:
Thanks for the info!

> I literally just discussed splitting this off into a separate module:

I think this would work perfectly here. The lack of any sitemap configuration method seems to imply that `myst_commonmark_only = True` can only be used for single file documentation, if I'm not mistaken, which is not great.

> note it is eventually intended for recommonmark to be deprecated by myst-parser

yeah wasn't really considering recommonmark, was just looking for ways in which myst could do similar things. MkDocs (and their material theme) I did actually consider after having admired FastAPI docs. In all its simplicity actually seems pretty great, but I guess the EBP people would kill me in my sleep if I go down that road :smile: 

In any other repo I'd happily use myst with all its syntax additions, but as I think I've mentioned before, I'd love to have it guaranteed that mdformat can self format all its own docs, using its git master branch with no plugins added. So this repo is a bit exceptional to me in that way.

chrisjsewell:
> I did actually consider after having admired FastAPI docs. In all its simplicity actually seems pretty great, but I guess the EBP people would kill me in my sleep if I go down that road 😄

😂 out of interest what would you say are the main selling points of FastAPI, i.e. is there anything you think myst should "borrow" from it

hukkinj1:
> out of interest what would you say are the main selling points of FastAPI

I think you mean MkDocs, not FastAPI? FastAPI just happens to have their docs built with MkDocs.

As a total noob, comparing Sphinx+MyST with MkDocs, what seems tempting about MkDocs to me is:
- Configuration is YAML, not Python source code. This feels "right" if my use case is simple and I couldn't care less about programming Sphinx. This probably makes it also a lot more approachable to non-Python projects.
- its built from the ground up with Markdown in mind. Sphinx docs focus on reST.
- [The material theme](https://github.com/squidfunk/mkdocs-material) (used by FastAPI among others) is great. Has dark mode. Has the dual pane navigation. There's a Sphinx port but that lacks dark mode for instance.
- a nice built-in docs server for development

There's also some icky sides to MkDocs like the fact that it parses John Gruber's Markdown instead of CommonMark, and that MyST isn't available (for any other repo but this I'd want the extra syntax).

Issue #224, opened 2021-06-17
Line Ending Changing with formatting


DragonCrafted87:
**Describe the bug**
File is always output with crlf line endings on windows even if no formatting was changed

**To Reproduce**

1. Start with a valid markdown file on Windows with lf(unix) line endings
2. run file through mdformat
3. file now has crlf(dos) line endings

**Expected behavior**
File has same line ending style it started with

**Environment**

Windows 10
Python 3.9.5
pre-commit 2.13.0
-   repo: https://github.com/executablebooks/mdformat
    rev: 0.7.6
    hooks:
    -   id: mdformat
        additional_dependencies:
        -   mdformat-gfm
        -   mdformat-black

hukkin:
Thanks for the issue!

To clarify, mdformat never creates `crlf`. It always normalizes end of lines to a single `lf` character. This is part of the whitespace normalization that mdformat does, and is by design.

If there's a good reason why one would want to use `crlf` instead, we can consider adding a `--crlf` flag. I'm not aware of a good use case, and allowing `crlf` adds some complexity (also in plugins etc.), so I haven't considered adding this.

Based on what you wrote, it seems you too may be happier with `lf`s than `crlf`s?

DragonCrafted87:
yeah, I do want `lf` endings which is why I created the bug I've created a repo with a minimal pre-commit setup that demonstrates what I mean https://github.com/DragonCrafted87/temp-test-repo.

I have the pre-commit-hooks mixed-line-ending hook set to make sure everything is `lf` followed by mdformat which is writing out changes and then if I run them again with `pre-commit run --all-files` both show changes again.

in the case of the repo I created I would expect no changes on either run as all I've done to the repo was had GitHub initializes it and add the pre-commit hook setup yet on the 2nd and following runs I get this for output:

Mixed line ending........................................................Failed
- hook id: mixed-line-ending
- exit code: 1
- files were modified by this hook

README.md: fixed mixed line endings

Check for added large files..............................................Passed
mdformat.................................................................Failed
- hook id: mdformat
- files were modified by this hook

hukkin:
I'm unable to reproduce on Linux:
```console
foo@bar:~$ git clone git@github.com:DragonCrafted87/temp-test-repo.git
Cloning into 'temp-test-repo'...
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (6/6), done.
[...]

PR #226, opened 2021-06-18
[ENH][FIX]  Add Configurable Line Endings

DragonCrafted87:
Fix for Issue #224 

Add new Parameter
--line-ending

that controls how the line endings are set for the output file

### DIFF
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -103,7 +103,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
                 )
                 return 1
             if path:
-                atomic_write(path, formatted_str, options.get("line-ending", None))
+                atomic_write(path, formatted_str, options.get("line_ending", None))
             else:
                 sys.stdout.write(formatted_str)
     if format_errors_found:
@@ -121,7 +121,7 @@ def validate_wrap_arg(value: str) -> Union[str, int]:
 
 
 def validate_line_ending_arg(value: str) -> str:
-    sys.stdout.write(value)
+    pprint(value)
     if value in {"\n", "\r", "\r\n"}:
         return value
     if value == "lf":
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -3,7 +3,6 @@ import contextlib
 import itertools
 import logging
 from pathlib import Path
-from pprint import pprint
 import shutil
 import sys
 import textwrap
@@ -58,7 +57,6 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
 
     # Convert args to a mapping
     options: Mapping[str, Any] = vars(args)
-    pprint(options)
 
     format_errors_found = False
     renderer_warning_printer = RendererWarningPrinter()
@@ -121,7 +119,6 @@ def validate_wrap_arg(value: str) -> Union[str, int]:
 
 
 def validate_line_ending_arg(value: str) -> str:
-    pprint(value)
     if value in {"\n", "\r", "\r\n"}:
         return value
     if value == "lf":
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -3,6 +3,7 @@ import contextlib
 import itertools
 import logging
 from pathlib import Path
+from pprint import pprint
 import shutil
 import sys
 import textwrap
@@ -57,7 +58,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
 
     # Convert args to a mapping
     options: Mapping[str, Any] = vars(args)
-    sys.stdout.write(options)
+    pprint(options)
 
     format_errors_found = False
     renderer_warning_printer = RendererWarningPrinter()
--- a/mdformat/_cli.py
+++ b/mdformat/_cli.py
@@ -119,6 +119,7 @@ def validate_wrap_arg(value: str) -> Union[str, int]:
 
 
 def validate_line_ending_arg(value: str) -> str:
+    print(value)
     if value in {"\n", "\r", "\r\n"}:
         return value
     if value == "lf":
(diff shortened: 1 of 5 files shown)


mdformat/_util.py
```diff
@@ -78,7 +78,7 @@ def is_md_equal(
     return html_texts["md1"] == html_texts["md2"]
 
 
-def atomic_write(path: Path, text: str) -> None:
+def atomic_write(path: Path, text: str, line_ending: Optional[str] = None) -> None:
```
hukkin, line 81:
```suggestion
def atomic_write(path: Path, text: str, line_ending: str) -> None:
```

mdformat/_util.py
```diff
@@ -87,7 +87,7 @@ def atomic_write(path: Path, text: str) -> None:
     """
     fd, tmp_path = tempfile.mkstemp(dir=path.parent)
```
hukkin, line 88:
```suggestion
    newline = "\r\n" if line_ending == "crlf" else "\n"
    fd, tmp_path = tempfile.mkstemp(dir=path.parent)
```

mdformat/_util.py
```diff
@@ -87,7 +87,7 @@ def atomic_write(path: Path, text: str) -> None:
...
     try:
-        with open(fd, "w", encoding="utf-8") as f:
+        with open(fd, "w", encoding="utf-8", newline=line_ending) as f:
```
hukkin, line 90:
```suggestion
        with open(fd, "w", encoding="utf-8", newline=newline) as f:
```

mdformat/_cli.py
```diff
@@ -118,6 +118,20 @@ def validate_wrap_arg(value: str) -> Union[str, int]:
...
+        return "\n"
+    if value == "cr":
+        return "\r"
+    if value == "crlf":
+        return "\r\n"
+    if value == "system":
+        return None
+    raise ValueError("line-ending must be either a specifier or valid line ending.")
+
+
```
hukkin, line 134:
```suggestion
```

mdformat/_cli.py
```diff
@@ -149,6 +163,13 @@ def make_arg_parser(
         metavar="{keep,no,INTEGER}",
         help="paragraph word wrap mode (default: keep)",
     )
+    parser.add_argument(
+        "--line-ending",
+        default="system",
+        type=validate_line_ending_arg,
+        metavar="{system,lf,cr,crlf}",
+        help="output file line endings (default: system)",
+    )
```
hukkin, line 172:
```suggestion
    parser.add_argument(
        "--end-of-line",
        default="lf",
        choices=("lf", "crlf"),
        help="output file line ending mode",
    )
```

mdformat/_cli.py
```diff
@@ -101,7 +101,7 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
                 )
                 return 1
             if path:
-                atomic_write(path, formatted_str)
+                atomic_write(path, formatted_str, options.get("line_ending", None))
```
hukkin, line 104:
```suggestion
                atomic_write(path, formatted_str, options["end_of_line"])
```

hukkin (changes requested):
Thank you _very __very__ much_ for doing the research and coming up with an excellent fix. Much appreciated.

I've requested some changes to reduce the amount of configuration as I don't want to add any API surface unless a real world user with the need for it appears.

In addition to these changes we need to fix this line: https://github.com/executablebooks/mdformat/blob/ee13da698bcb24c1af05bf243b341845e4b60bdc/mdformat/_api.py#L62

We need to change it to 
```python
atomic_write(f, formatted_md, options.get("end_of_line", "lf"))
```

Also we need a simple test that ensures no `crlf`s are written to files by default. Running that in CI on Windows will validate our changes and prevent regressions. I can also add this test myself if you're not familiar with the test setup.

hukkin:
I'm actually a bit tempted to not add the configuration option at all, and just always use a `\n` character always. This is what I thought the current behavior was until you proved otherwise :smile: 

I'm not sure if there's a good reason to prefer `\r\n` and if there's one person on earth who needs that, running in serial with an end of line fixer should solve that.

If it's later apparent that many people need `\r\n` then the configuration could be introduced.

What do you think @DragonCrafted87 ?

DragonCrafted87:
yeah as while hardcoding it to purely `lf` would work in my case and I imagine most others I'm sure there's someone who wants all `crlf` line endings and like you said it will fail if there is an end of line fixer running in line with it in the same way I noticed it. 
I should be able to clean up the rest of the comments this evening though I may need a hand with the test as I tend to use python more for scripting than anything complicated enough to need a test suite

hukkin:
:+1: 

For tests, please add these to `tests/test_cli.py`:
```python
def test_eol__lf(tmp_path):
    file_path = tmp_path / "test.md"
    file_path.write_bytes(b"Oi\r\n")
    assert run([str(file_path), "--end-of-line=lf"]) == 0
    assert file_path.read_bytes() == b"Oi\n"


def test_eol__crlf(tmp_path):
    file_path = tmp_path / "test.md"
    file_path.write_bytes(b"Oi\n")
    assert run([str(file_path), "--end-of-line=crlf"]) == 0
    assert file_path.read_bytes() == b"Oi\r\n"
```

and these to `tests/test_api.py`:
```python
def test_eol__lf(tmp_path):
    file_path = tmp_path / "test.md"
    file_path.write_bytes(b"Oi\r\n")
    mdformat.file(str(file_path), options={"end_of_line": "lf"})
    assert file_path.read_bytes() == b"Oi\n"


def test_eol__crlf(tmp_path):
    file_path = tmp_path / "test.md"
    file_path.write_bytes(b"Oi\n")
    mdformat.file(str(file_path), options={"end_of_line": "crlf"})
    assert file_path.read_bytes() == b"Oi\r\n"
```

DragonCrafted87:
that push should have everything accounted for to just do the `crlf` and `lf` options. Thank you for the tests if I ever wind up doing a larger project in python I'll have to take a closer look at pytest as it looks like its nice and straight forward for a test system

hukkin:
Thanks this is great! I'll try to have this released later today.

Issue #317, opened 2022-03-16
Indentation with 3 spaces breaks mkdocs


ThmX:
### Describe the bug

**context**
When I format my lists markdown using mdformat.

**expectation**
I expected the file to be correctly formatted with mkdocs.

**bug**
But instead the lists are merged due to the indentation with 3 spaces.

**problem**
This is a problem for people using mkdocs because files are not formatted as expected.
The documentation from mkdocs states that it needs 4 spaces as indentation:
<https://www.mkdocs.org/user-guide/writing-your-docs/#:~:text=If%20a%20line%20is%20indented%20by%204>

### Reproduce the bug

A list of lists with 4 spaces as indent is correctly displayed in mkdocs 
```
- Some list
    - A
    - B
    - C
- Something else
```

Which is then reformatted with 3 spaces and then both list are merged into one by mkdocs
```
- Some list
   - A
   - B
   - C
- Something else
```

### List your environment

❯ python --version
Python 3.9.10

❯ python -m mdformat --version
mdformat 0.7.14 (mdformat_myst: 0.1.5, mdformat_tables: 0.4.1, mdformat_frontmatter: 0.4.1, mdformat_web: 0.1.0, mdformat_black: 0.1.1, mdformat_beautysh: 0.1.1,
mdformat_config: 0.1.3)

hukkin:
Thanks for the issue!

Firstly, I'd like to clarify that mdformat uses 2 spaces, not 3 spaces as unordered list item indentation.

Secondly, I assume you refer to this piece of text in mkdocs docs:
> If a line is indented by 4 or more spaces, that line is assumed to be an additional line of the value for the previous keyword. A keyword may have as many lines as desired. All lines are joined into a single string.

That text has nothing to do with list item indentation. It is under a section that describes "MultiMarkdown Style Meta-Data", which is syntax that CommonMark is unaware of, so mdformat will not respect it unless you create a plugin supporting it.

ThmX:
Oh you are right, I read too fast :-/

However, the issue is still there. I cannot find the minimum indentation width in the documentation from `mkdocs`, but from what I see in my documentation, it needs 4 spaces to be able to format correctly. Otherwise it simply flatten all the lists.

Plus I must admit I like the indentation with 4 spaces (same as in PyCharm basically)

I created PR #318 as a fix by adding an option to set the minimum indentation width.

What do you think?

Not sure if I could create it as a plugin instead (also if it makes sense having this as a plugin)

[...]

PR #318, opened 2022-03-16
Add an option to set the minimum indentation spaces

ThmX:
Fixes https://github.com/executablebooks/mdformat/issues/317


hukkin:
I appreciate the effort! I think that with KyleKing's `mdformat-mkdocs` plugin this is not necessary. Closing.

PR #333, opened 2022-07-01
renderers as parameters

inktrap:
I would like to add a plaintext renderer that works out of the box with all the plugins/extensions. I would like a generic render_tree function for this (which the current one is not, as it adds newlines -- but that is alright with me).

My approach would be to add renderers as a parameter, so other projects can provide a different set of default rules. Alternatively one could set the default to `None` and then assign `DEFAULT_RENDERERS` later. While this is cleaner, the type signature would have to account for this `Union[Mapping[str, Render], NoneType]`.

Alternatively I would duplicate the whole function or would write a renderer class with methods for all the token types of all the plugins.

### DIFF
--- a/src/mdformat/renderer/__init__.py
+++ b/src/mdformat/renderer/__init__.py
@@ -19,7 +19,7 @@ from markdown_it.token import Token
 
 from mdformat.renderer._context import DEFAULT_RENDERERS, WRAP_POINT, RenderContext
 from mdformat.renderer._tree import RenderTreeNode
-from mdformat.renderer.typing import Postprocess
+from mdformat.renderer.typing import Postprocess, Render
 
 LOGGER = logging.getLogger(__name__)
 
--- a/src/mdformat/renderer/__init__.py
+++ b/src/mdformat/renderer/__init__.py
@@ -62,6 +62,7 @@ class MDRenderer:
         options: Mapping[str, Any],
         env: MutableMapping,
         *,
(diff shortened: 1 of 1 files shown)


hukkin:
Hi thanks for the PR!

Firstly, I'm somewhat a minimalist when it comes to public API and don't want to expand it unless strictly required. It's not clear to me how this API expansion is valuable to upstream.

A few questions come to mind.
1. If you need a plain text renderer, are you sure that Markdown renderer is a good superclass to begin with?
2. If `render_tree` needs these changes for your use case and you don't like the newlines it creates, why don't you override it?

inktrap:
1.1. I liked the idea that the actual rules are defined in `_context.py` and that it is separated from the application/functional aspects, so I don't have to duplicate that e.g. rendering a tree, applying postprocessors and plugins, … but it is not completely separated and I had the impression that it is better to de-couple this. I get that it works coupled for mdformat.
1.2. I have a similar wish regarding `text()` that (if i remember correctly) constructs a markdown-it-py instance and does both, parsing and rendering, while I would prefer it separately, e.g. create a function `token()`, pass it a list of `Token` objects, let `text()` do the parsing and let it call `token()`.
2. I only need minimal changes and didn't want to copy/paste the whole thing. I would have to keep it updated and I thought that being generic would be a better idea. Sure, I can overwrite it. The function of mdformat is to provide markdown rendering, not a framework for other uses-cases.

inktrap:
I simply implemented the renderer as a class, without using mdformat.

Issue #336, opened 2022-07-16
Add `--end-of-line=native`


marktsuchida:
### Context

I have a repo that uses mdformat as a pre-commit hook; it works beautifully on Linux and macOS.

If I clone the repo on Windows, under Git's default settings (`core.autocrlf` = `true`), the Markdown files in the working tree have CRLF newlines. When the pre-commit hook runs, it replaces newlines with LF (as expected in the absence of `--end-of-line`). This leads to the confusing situation where:
- pre-commit reports no errors, yet
- after the commit, there remain unstaged changes (consisting of CRLF -> LF replacement).

Specifying `--end-of-line=crlf` can fix this, but we would need to specify this on Windows only, and there is no way to do so in `.pre-commit-config.yaml` or `.mdformat.toml`, as far as I can tell.

It is probably also possible to work around this using `.gitattributes`, but that adds more things to maintain.

### Proposal

- Add the choice `native`, in addition to the current choices (`lf`, `crlf`), to the `--end-of-line` option.
    - And similarly to the `end_of_line` key of `.mdformat.toml`.
- When `--end-of-line=native` is given, use the value of `os.linesep` to determine the newlines to use.
- The default newline remains `lf` to maintain compatibility.

The choice of the name `native` is bikesheddable (perhaps `auto`?).

### Tasks and updates

- [x] Settle on a name (`native`, `auto`, or other)
- [ ] Implement

marktsuchida:
One reason to prefer `native` over `auto` is that some people may think that `auto` means keep the existing newline style of the file, which is not what I'm proposing here (though also a possible feature).

hukkin:
Thanks for the issue!

My understanding is that `core.autocrlf` is set to `false` by default (see https://stackoverflow.com/a/48230871). Would it solve your issue if you change your local git configuration to that? My understanding is that most modern text editors and IDEs on Windows work fine with bare LF newlines, so `autocrlf` is probably not needed.

marktsuchida:
Thanks for the quick reply.

The Windows installer for Git, by default, sets the effective default to `core.autocrlf` = `true` (see the second [answer](https://stackoverflow.com/a/49207740) to the same SO question; I also checked this myself today), so most people will have that setting by default on Windows. And personally I think it is a wise default.

[...]

PR #345, opened 2022-08-22
add option `end-of-line: keep`

Cube707:
as discussed in #336 

The new option option `--end-of-line=keep` will not attempt to change line-endings, instead it will infer the type of lineedning from the first line of the file and continue using that.

---

not jet complete, it failes on `crlf` lineendings when using `--check`, as the `formated_str` will have the new-line-characters replaced and not equal `original_str`, therefor "correctly" failing. I could of course skip the `opts["check"]` block, but that feels really wrong.

any suggestions?

### DIFF
--- a/docs/users/configuration_file.md
+++ b/docs/users/configuration_file.md
@@ -19,5 +19,5 @@ Command line interface arguments take precedence over the configuration file.
 #
 wrap = "keep"       # possible values: {"keep", "no", INTEGER}
 number = false      # possible values: {false, true}
-end_of_line = "lf"  # possible values: {"lf", "crlf"}
+end_of_line = "lf"  # possible values: {"lf", "crlf", "keep"}
 ```
--- a/src/mdformat/_api.py
+++ b/src/mdformat/_api.py
@@ -7,7 +7,13 @@ from pathlib import Path
 from typing import Any
 
 from mdformat._conf import DEFAULT_OPTS
-from mdformat._util import EMPTY_MAP, NULL_CTX, atomic_write, build_mdit
+from mdformat._util import (
+    EMPTY_MAP,
+    NULL_CTX,
+    atomic_write,
+    build_mdit,
+    detect_newline_type,
+)
 from mdformat.renderer import MDRenderer
 
 
@@ -58,7 +64,7 @@ def file(
     if f.is_symlink():
         raise ValueError(f'Cannot format "{f}". It is a symlink.')
 
-    original_md = f.read_text(encoding="utf-8")
+    original_md = f.read_bytes().decode()
     formatted_md = text(
         original_md,
         options=options,
@@ -66,9 +72,7 @@ def file(
         codeformatters=codeformatters,
         _filename=str(f),
     )
-    newline = (
-        "\r\n"
-        if options.get("end_of_line", DEFAULT_OPTS["end_of_line"]) == "crlf"
-        else "\n"
+    newline = detect_newline_type(
+        original_md, options.get("end_of_line", DEFAULT_OPTS["end_of_line"])
     )
     atomic_write(f, formatted_md, newline)
--- a/src/mdformat/_util.py
+++ b/src/mdformat/_util.py
@@ -117,7 +117,7 @@ def atomic_write(path: Path, text: str, newline: str) -> None:
         raise
 
 
-def detect_newline_type(md: str, eol_setting) -> Literal["\n", "\r\n"]:
+def detect_newline_type(md: str, eol_setting: str) -> Literal["\n", "\r\n"]:
     """Returns the newline-character to be used for output.
 
     If `eol_setting == "keep"`, the newline character used in the passed
--- a/src/mdformat/_util.py
+++ b/src/mdformat/_util.py
@@ -127,10 +127,7 @@ def detect_newline_type(md: str, eol_setting: str) -> Literal["\n", "\r\n"]:
     """
     if eol_setting == "keep":
         first_eol = RE_NEWLINES.search(md)
-        if first_eol is None:
-            return "\n"
-        char = first_eol.group()
-        return "\n" if char == "\r" else char
+        return "\r\n" if first_eol and first_eol.group() == "\r\n" else "\n"
     if eol_setting == "crlf":
         return "\r\n"
     return "\n"
--- a/src/mdformat/_util.py
+++ b/src/mdformat/_util.py
@@ -7,7 +7,8 @@ from pathlib import Path
 import re
 import tempfile
 from types import MappingProxyType
-from typing import Any, Literal
+from typing import Any
+from mdformat._compat import Literal
 
 from markdown_it import MarkdownIt
 from markdown_it.renderer import RendererHTML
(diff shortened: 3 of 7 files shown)


docs/users/configuration_file.md
```diff
@@ -19,5 +19,5 @@ Command line interface arguments take precedence over the configuration file.
 #
 wrap = "keep"       # possible values: {"keep", "no", INTEGER}
 number = false      # possible values: {false, true}
-end_of_line = "lf"  # possible values: {"lf", "crlf"}
+end_of_line = "lf"  # possible values: {"lf", "crlf", "keep}
```
hukkin, line 22:
```suggestion
end_of_line = "lf"  # possible values: {"lf", "crlf", "keep"}
```

src/mdformat/_api.py
```diff
@@ -68,7 +68,11 @@ def file(
...
     newline = (
         "\r\n"
-        if options.get("end_of_line", DEFAULT_OPTS["end_of_line"]) == "crlf"
+        if (
+            options.get("end_of_line", DEFAULT_OPTS["end_of_line"]) == "crlf"
+            or options.get("end_of_line", DEFAULT_OPTS["end_of_line"]) == "crlf"
+            and original_md[original_md.find("\n") - 1] == "\r"
+        )
         else "\n"
     )
```
hukkin, line 77: I think the conditional expression is getting too long to be readable here. I suggest replacing this with
```python
eol_setting = options.get("end_of_line", DEFAULT_OPTS["end_of_line"])
if eol_setting == "keep":
    first_eol = re.search(r"\r\n|\r|\n", original_str)
    newline = "\r\n" if first_eol and first_eol.group() == "\r\n" else "\n"
elif eol_setting == "crlf":
    newline = "\r\n"
else:
    newline = "\n"
```

And maybe also move this to a function (that takes `eol_setting` string as input) and share the code with `_cli.py`

src/mdformat/_api.py
```diff
@@ -68,7 +68,11 @@ def file(
...
```
hukkin, line 68: Can't comment on the exact line, but if you look above, on line 61 we have
```python
original_md = f.read_text(encoding="utf-8")
```
We need to change this to
```python
original_md = f.read_bytes().decode()
```
in order to preserve line ending sequences.

hukkin (changes requested):

hukkin:
> I could of course skip the opts["check"] block, but that feels really wrong.

Yeah we shouldn't skip this. What we need to do is check which newline sequence `"keep"` resolves to. If it resolves to `\n` then execute the code black as if `opts["end_of_line"] == "lf"` and if it resolves to `\r\n` execute the code block as if `opts["end_of_line"] == "crlf"`

I realize this can be a bit unclear explanation :smile: If so, please ask for details or I can write the implementation for you.

Cube707:
> I realize this can be a bit unclear explanation

I absoloutly understand what you mean, but further investigation revealt that there is an existing issue that makes this fail even if `keep` would jump into the correct codeblocks (see #347)

docs/users/configuration_file.md
Cube707: ups 🤣

src/mdformat/_api.py
Cube707: yea, externalising this makes sense. I also have to think about this again now that we defnetly want to upport `\r`.

Cube707: also sidenote: it would probably be sensible to rewrite this at some point so that the cli uses the api code and not having two version of nearly identical code. But thats a lot of work for some other time

hukkin: :+1: The code block I provided should work fine with `\r`. It'll convert them to `\n` as per the specification in https://github.com/executablebooks/mdformat/issues/336#issuecomment-1224883207

Cube707: I also just read it completely and realsied you did more than just untangle the `if`

I also tested `first_eol = re.search(r"\r\n|\r|\n", original_str)` and found something wird with it, but can't remember what. I will retest and come back to this

src/mdformat/_api.py
Cube707: I will check this.

I also realsied I forget tests for the API side in generall.. 🤦🏼‍♂️

src/mdformat/_util.py
```diff
@@ -7,7 +7,7 @@
 import re
 import tempfile
 from types import MappingProxyType
-from typing import Any
+from typing import Any, Literal
```
hukkin, line 10:
```suggestion
from typing import Any
from mdformat._compat import Literal
```

src/mdformat/_util.py
```diff
@@ -114,3 +115,21 @@ def atomic_write(path: Path, text: str, newline: str) -> None:
     except BaseException:  # pragma: no cover
         os.remove(tmp_path)
         raise
+
+
+def detect_newline_type(md: str, eol_setting) -> Literal["\n", "\r\n"]:
```
hukkin, line 120:
```suggestion
def detect_newline_type(md: str, eol_setting: str) -> Literal["\n", "\r\n"]:
```

src/mdformat/_util.py
```diff
@@ -114,3 +115,21 @@ def atomic_write(path: Path, text: str, newline: str) -> None:
...
+    """Retruns the newline-character to be used for output.
```
hukkin, line 121:
```suggestion
    """Returns the newline-character to be used for output.
```

src/mdformat/_util.py
```diff
@@ -114,3 +115,21 @@ def atomic_write(path: Path, text: str, newline: str) -> None:
...
+
+    If `eol_setting`=="keep", the newline character used in the passed
```
hukkin, line 123:
```suggestion
    If `eol_setting == "keep"`, the newline character used in the passed
```

hukkin (changes requested):

src/mdformat/_util.py
Cube707: ups

tests/test_cli.py
```diff
@@ -278,6 +292,18 @@ def test_eol__check_crlf(tmp_path):
     assert run((str(file_path), "--check", "--end-of-line=crlf")) == 0
 
 
+def test_eol__check_keep_lf(tmp_path):
+    file_path = tmp_path / "test.md"
+    file_path.write_bytes(b"lol\n")
+    assert run((str(file_path), "--check", "--end-of-line=keep")) == 0
```
hukkin, line 298: I think there's work to be done in the `if opts["check"]:` block.

Try adding this test. I suspect it will fail right now until the aforementioned is fixed
```suggestion

    file_path.write_bytes(b"lol\n")
    assert run((str(file_path), "--check", "--end-of-line=keep")) == 0
    
    file_path.write_bytes(b"mixed\nEOLs\r")
    assert run((str(file_path), "--check", "--end-of-line=keep")) == 1
```

hukkin: I've committed this and the fix to `if opts["check"]:`.

I'll release this today.

hukkin (approved):

hukkin:
This seems great! Thanks for your contribution @Cube707 !

Issue #347, opened 2022-08-24
`--check` doesn't work with `--end-of-line=crlf`


Cube707:
### Describe the bug

**context**
running `mdformat --check --end-of-line=crlf` on a correctly formated markdonw file result in error.

**expectation**
I expected `mdformat` to succede and nothing to happen

**bug**
But instead an error is rased that the file is not formated:

```console
$ mdformat --check --end-of-line=crlf test.md
Error: File "D:\Downloads\temp\mdformat\test.md" is not formatted.
```

**related code**

I belive the problem thems from here:

https://github.com/executablebooks/mdformat/blob/24f9d25960d28efc75cbfa051881bac5fc46a842/src/mdformat/_cli.py#L85-L87

debugging reveald the content of `formatted_str` and `original_str` to have different lineendings:

```
original_str: '# Test file\r\n\r\nHello world\r\n'
formatted_str: '# Test file\n\nHello world\n'
```

I assume that the formating process internally does a [_universal newline_](https://docs.python.org/3/glossary.html#term-universal-newlines) conversion, resulting in all newline types beeing converted to `\n`

### Reproduce the bug

1. create a well formated mardkown file `test.md` with dos-style line endings (`\r\n`):

   ```markdown
   # Test file
   
   Hello world
   
   ```
2. run mdformat on it with the `--check` and `--end-of-line=crlf` options:

   ```bash
   mdformat --check --end-of-line=crlf test.md
   ```

3. get error

### List your environment

OS: windows 10
Python: v3.10.6
mdformat: 0.7.15

hukkin:
Thanks for the issue!

Yeah this seems to need better tests and maybe rethink the whole `if opts["check"]: ` block.

Cube707:
I belive this could benifit from rethingking the whole _line-endings_ proccedure. We should probaly do something like this:

* read the markdown-file with no lineending conversion applyed (`open(file, "r", newline="")`)
* figure out and store the first newline character
* apply _universal newline_ conversion by wrapping the text in a second layer of `StringIO` 
* pass the content on, now that it matches python conventions and should be what `markdown_it` expects and returns
* write out the formated file, providing the newline character as per the options

proof of concept:

```python
from io import StringIO


with open("test.md", "r", newline="") as fp:
    firstline = fp.readline()
    if firstline[-2:]=="\r\n":
        newlinechar = "\r\n"
    elif firstline[-1:]=="\r":
        newlinechar = "\r"
    else:
[...]

PR #348, opened 2022-08-24
FIX: `--check` not working with CRLF line endings

hukkin:
Fixes https://github.com/executablebooks/mdformat/issues/347


Cube707:
This nukes any posibily to detect and fail on mixed lineendings if I am not mistakes?

Cube707:
> ~~This nukes any posibily to detect and fail on mixed lineendings if I am not mistakes?~~

never mind, the next lines still handle that

PR #358, opened 2022-09-14
Do not touch the original file if the result is identical to the original

### LINKED ISSUE #357: Do not update the timestamp of the files when they are left unchanged
### Describe the bug

mdformat updates the timestamp of the files even when they are left unchanged. Instead, mdformat should not touch files when they are left unchanged (as `black` does for example).

### Reproduce the bug

Use stat and mdformat on one well formatted file.

### List your environment

_No response_

paugier:
Fixes #357


hukkin:
Thanks, this is great!

I added a couple tests and merged.

Issue #353, opened 2022-08-29
mdformat failure when EM SPACE character is present


nschloe:
### Describe the bug

When running mdformat on
````markdown
```
a
```
 
# A
````
(`"'```\\na\\n```\\n\\u2003\\n# A\\n'"`) I'm getting an error. Might be related to https://github.com/executablebooks/mdformat/issues/351.

### Reproduce the bug

See above.

PR #364, opened 2022-10-11
Compare parsed HTML

hukkin:


hukkin:
Note to self: This PR was made originally to fix https://github.com/executablebooks/mdformat/issues/353, and to help with any similar future issues. https://github.com/executablebooks/mdformat/issues/353 is now fixed though.

If no good reason comes up to do what this PR does, merging this should be avoided, as this introduces pure Python HTML parsing and a lot of object creation to every file format operation. Current naive HTML string comparison (even if with a few regex replacements) should be a lot faster.

PR #375, opened 2022-12-18
DOCS: link to the mdformat-admon plugin

KyleKing:
I'm not sure if you're open to contributions here, but I thought adding a link to my new plugin might be useful!

### DIFF
--- a/docs/users/plugins.md
+++ b/docs/users/plugins.md
@@ -130,7 +130,7 @@ formatted = mdformat.text(unformatted, extensions={"tables"})
   </tr>
   <tr>
     <td><a href="https://github.com/KyleKing/mdformat-admon">mdformat-admon</a></td>
-    <td><code>admonitions</code></td>
+    <td><code>admonition</code></td>
     <td>Adds support for <a href="https://python-markdown.github.io/extensions/admonition/">python-markdown</a> admonitions</td>
   </tr>
 </table>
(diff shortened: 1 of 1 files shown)


docs/users/plugins.md
```diff
@@ -128,4 +128,9 @@ formatted = mdformat.text(unformatted, extensions={"tables"})
     <td><code>deflist</code></td>
     <td>Adds support for <a href="https://pandoc.org/MANUAL.html#definition-lists">Pandoc-style</a> definition lists</td>
   </tr>
+  <tr>
+    <td><a href="https://github.com/KyleKing/mdformat-admon">mdformat-admon</a></td>
+    <td><code>admonitions</code></td>
```
KyleKing, line 133: Should the rows in this table be sorted alphabetically or grouped?

hukkin, line 133:
```suggestion
    <td><code>admonition</code></td>
```
I think this should match the string that is used here https://github.com/KyleKing/mdformat-admon/blob/2b55fb462fac5e4e6b1abfad1f3d27c75bd1209c/pyproject.toml#L36

hukkin: I think this used to be in alphabetical order but no longer is, so I don't care much. The order can be fixed seprately for all items if seen necessary.

hukkin (changes requested):
Thanks, this is very welcome addition to the docs!

docs/users/plugins.md
KyleKing: Thanks! I'm happy to make that change this evening, or you can go ahead and merge

Issue #268, opened 2021-09-17
Add '--diff' flag to show changes without modifying the source file


jamesquilty:
### Description / Summary

A nice-to-have feature would be for mdformat to show the changes it would make to the file by writing the output to stdout rather than to the original file. This is a feature that [black](https://github.com/psf/black) has, using the `--diff` flag on the command line.

### Value / benefit

At present, the only way to find out the changes that `mdformat` will make (or has made) is to create a copy of the file to be formatted, run `mdformat` on the copy and compare the two files. Or, if the file is under version control, to compare to the previous version (e.g. by `git diff`). The extra effort is a bit of a pest, but this feature is really just a nice-to-have.

hukkin:
Thanks for the issue!

I'm not _too_ excited about this feature as
- it's pretty easy to print the diff by doing something like the following already
  ```sh
  diff foo.md <(mdformat - < foo.md)
  ```
- I find it more useful to know if a file is formatted or not, rather than what exact parts are not formatted. The best way to fix is by running `mdformat` anyways, I don't need concern myself with the formatting details.

I'll leave this open for now so we can reconsider if there's demand and real-life use cases from other users.

jamesquilty:
Well, I'm not _too_ excited about the `--check` feature for exactly the same reasoning. YMMV. Black's implementation of this feature appears to be straightforward and I've added it to the **Implementation details** section above.

psychemedia:
In passing, I note that `nbqa` [now supports](https://github.com/nbQA-dev/nbQA/issues/555#issuecomment-974651504) checks such as running `mdformat` over notebook Markdown cells, including the running of an `--nbqa-diff` switch that generates a preview rather than changing content in-place.

veghdev:
I think a real use case could be that people submit a PR for documents in Markdown files via the Github web editor. If they make an error, CI will fail, but there is no way to tell the user what the formatting problem is. In many cases these people cannot use a development environment, but it is possible to fix their PRs with the diff output.

MrMino:
@hukkin 

> it's pretty easy to print the diff by doing something like the following already

That assumes you are using 1) a shell 2) one that has the process substitution feature. Neither is the case if you are running it via `exec` call on a Docker container that uses Ash for its shell, i.e. in many CI flavors out there.

Lots of formatters have this feature exactly for this reason: so that there's an easy way to show the diff in CI logs.

hukkin:
[...]

PR #381, opened 2023-01-26
Add --diff option to show diff when running with --check

tdgroot:
We use `mdformat` to validate Markdown formatting in a [documentation project](https://github.com/ByteInternet/hypernode-docs-next) we're running. Most users are contributing to the documentation by creating changes, commits and pull requests from the Github UI, so they won't be running any linting/validation on commit. In this case, it would be very useful for those users to see what `mdformat` is marking as "unformatted" and knowing what should be done to fix that.

This was highly inspired by the implementation that [Black](https://github.com/psf/black) uses, as it seems to doing just what we need.

Looking forward to your thoughts and review :).

### DIFF
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 from collections.abc import Callable, Generator, Iterable, Mapping, Sequence
 import contextlib
+from datetime import datetime
 import itertools
 import logging
 from pathlib import Path
@@ -17,6 +18,7 @@ from mdformat._conf import DEFAULT_OPTS, InvalidConfError, read_toml_opts
 from mdformat._util import atomic_write, detect_newline_type, is_md_equal
 import mdformat.plugins
 import mdformat.renderer
+from src.mdformat._output import diff
 
 # Match "\r" and "\n" characters that are not part of a "\r\n" sequence
 RE_NON_CRLF_LINE_END = re.compile(r"(?:[^\r]|^)\n|\r(?:[^\n]|\Z)")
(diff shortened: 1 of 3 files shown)


AlexanderGrooff (approved):

AlexanderGrooff:
Would be great to have this feature! :rocket:

tdgroot:
@hukkin any chance to get a review for this PR? :)

hukkin:
Hi @tdgroot, thanks for contributing and sorry for the long wait!

> Most users are contributing to the documentation by creating changes, commits and pull requests from the Github UI, so they won't be running any linting/validation on commit. In this case, it would be very useful for those users to see what mdformat is marking as "unformatted" and knowing what should be done to fix that.

I think you may find setting mdformat up with https://pre-commit.ci/ and pre-commit even more useful than this feature. The fixes would be automatically applied to PRs -> the contributor doesn't have to care about failures or bad formatting at all.

Regardless, I've decided to accept this feature. Do you still want to work on this or should I continue from where you left?

We should make `--diff` work also without `--check`, returning 0 also in the case there is a diff. Also mypy compains, and test coverage is not 100%.

hukkin:
Closes https://github.com/executablebooks/mdformat/issues/268

tdgroot:
Hi @hukkin,

That's amazing! I'm not available to work on it this week, but I can update this PR next week I think. If that's taking too long for you, feel free to take my work and finish it :).

Kristinita:
@tdgroot, @hukkin, is there any chances that the `--diff` will still be added to Mdformat?

Thanks.

tdgroot:
@hukkin @Kristinita I finally updated the PR :sweat_smile:

What do you think?

PR #391, opened 2023-04-04
Add exclude option to cli and configuration toml

### LINKED ISSUE #359: Feature request: exclude/ignore option
### Context

Hello!

I would love to use it as `mdformat .`  (like black)  and have it ignore all files I know should be ignored and include any that should be included but is not explicitly formatted with .md extension.

I understand that the current suggested usage as a pre-commit hook uses their exclusion argument to exclude files. But i feel like having it in the cli would be great.

Let me know what you think.

### Proposal

I would like to propose adding exclude and include options to the cli (and by extension the configuration toml).
I believe it would be great for its usage in the command line.

### Tasks and updates

_No response_

thetawom:
PR adds an `exclude` option to the CLI (and to the configuration TOML) for excluding files matching the specified patterns. This provides a way to run `mdformat` outside of pre-commit but still with the ability to exclude certain files.

Resolves #359

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
   hooks:
   - id: absolufy-imports
 - repo: https://github.com/PyCQA/isort
-  rev: c5e8fa75dda5f764d20f66a215d71c21cfa198e1  # frozen: 5.10.1
+  rev: e44834b7b294701f596c9118d6c370f86671a50d  # frozen: 5.12.0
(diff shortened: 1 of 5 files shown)


twmht:
This really helps! thank you @thetawom

hukkin:
Thanks, I appreciate this! Closing in favor of https://github.com/executablebooks/mdformat/pull/451 which uses recursive globs and matches against relative paths.

Issue #383, opened 2023-02-07
mdformat._util.atomic_write fails if dir isn't writeable


guykisel:
### Describe the bug

**context**
I'm running mdformat on a directory of markdown docs in a perforce repo. In perforce, everything in a repo defaults to read-only, and only becomes writeable when checked out. Normally this should be fine, because I'm running mdformat on files I've checked out. However, the atomic_write function creates a temp dir in the current path, which in this case is not writeable unless I've explicitly checked it out.

**expectation**
I expect atomic_write to fall back to creating the temp dir in one of the default paths used by `tempfile.gettempdir()` as described at https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir or perhaps alternatively to give up on atomic writes and just do a best-effort?

**bug**
atomic_write fails with a permission error :(

```console
> mdformat .
Traceback (most recent call last):
  File "D:\prj\python\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\docs\Scripts\mdformat.exe\__main__.py", line 7, in <module>
  File "C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\docs\lib\site-packages\mdformat\__main__.py", line 8, in run
    exit_code = mdformat._cli.run(sys.argv[1:])
  File "C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\docs\lib\site-packages\mdformat\_cli.py", line 114, in run
    atomic_write(path, formatted_str, newline)
  File "C:\Users\user\AppData\Local\pypoetry\Cache\virtualenvs\docs\lib\site-packages\mdformat\_util.py", line 115, in atomic_write
    os.replace(tmp_path, path)
PermissionError: [WinError 5] Access is denied: 'D:\\prj\\docs\\tmpyw2fgxhp' -> 'D:\\prj\\docs\\README.md'
Error: Sequence aborted after failed subtask 'mdformat'
```

**problem**
This makes using mdformat with perforce slightly more difficult. It's not a blocker, just mildly annoying.

### Reproduce the bug

1. Set up a perforce repo
2. Create a dir with markdown files in the repo
3. Check out some of the files in a changelist, but not the parent dir
4. Run mdformat on the files

### List your environment

Python 3.9.1

```
> python -m pip freeze
attrs==22.2.0
CacheControl==0.12.11
certifi==2022.12.7
[...]

PR #404, opened 2023-07-14
fix: No permissions to create file in mountpoint under Docker

mihaigalos:
Fixes https://github.com/executablebooks/mdformat/issues/383 when dir isn't writable (i.e.: in a docker mount).


hukkin:
Thank you! https://github.com/executablebooks/mdformat/pull/409 fixes this and other reported issues.

Issue #376, opened 2022-12-29
Document a new plugin for shorter thematic breaks


csala:
### Context

I am using `mdformat` but having troubles with the current thematic break rendering composed of 70 underscore characters, which I am rather used to typing as three dashes `---`.

My (very personal and opinionated) reasons being:
1. The default `vim` highlighting for Markdown syntax breaks when multiple underscores are put together, because it conflicts with the italics format. Notice that this can probably be solved with Vim plugins or configuration tweaks, but the origin of the problem is about raw `vim` with no plugins.
2. Typing `---` while writing a document is quick and easy, while typing a line of exactly 70 underscores is comparatively much harder and needs some sort of copy/paste action. My usual coding workflow involves code style checkers like `mdformat` and `black` as `pre-commit` hooks, but in general I try to type in code and documentation that is already valid to minimize the number of times that these tools have to modify what I have typed, which forces me to redo the review, `git add` and `pre-commit` loop.

I admit that these are very personal reasons, so I do not wish to engange in any discussions about [the current rendering  style and the rationale behind it](https://mdformat.readthedocs.io/en/stable/users/style.html#thematic-breaks).

Instead, I thought best to take the route of coding a new plugin which overrides the rendering style for thematic breaks with three dashes, so anyone who wants can take advantage of it without having to do any change in the main `mdformat` project: https://github.com/csala/mdformat-simple-breaks

### Proposal

The proposal is just to add the plugin to the corresponding [list in the documentation](https://mdformat.readthedocs.io/en/stable/users/plugins.html#parser-extension-plugins)
I'm happy to make a PR for it.

### Tasks and updates

- [ ] Discuss and decide if a PR is wanted
- [ ] Do the corresponding PR

hukkin:
A PR is very welcome! :smile_cat:

hukkin:
Closing (https://github.com/executablebooks/mdformat/pull/418)

PR #418, opened 2023-12-09
docs: add plugins mdformat-pyproject and mdformat-simple-breaks

### LINKED ISSUE #379: Document new `mdformat-pyproject` plugin
### Context

I created [this plugin](https://github.com/csala/mdformat-pyproject) that lets `mdformat` load options from a `[tool.mdformat]` section inside the `pyproject.toml` file, if it exists. This basically covers the need exposed in #356 without having to make any modification to the core of `mdformat` and therefore not binding the project to any particular project type.

The `pyproject.toml` file is found relative to each file that is processed, and relative to `cwd` if `stdin` contents are being parsed. This is the same behavior as with `.mdformat.toml`.

Also, if `.mdformat.toml` exists, its options are also loaded, giving precedence to the `pyproject.toml` ones if an option exists in both places (basically, `.mdformat.toml` is loaded first as usual, and then the `pyproject.toml` update whatever options were already loaded).

### About the implementation

The implementation is slightly hacky, because it imports `mdformat` internals and forces the parsing and validation of CLI arguments twice in order to honor them, but I could not figure out a better way to achieve this without including some modification in `mdformat` to keep track of what the original CLI args were.

@hukkin If there is a cleaner or most robust way to achieve the same results I will happily hear that and adapt. Maybe #378 can also be considered along the way, so we can basically unify how the options and CLI arguments are handled in away that lets the plugins alter that if necessary (to either add options, or load the options from somewhere else).

### Proposal

Add this plugin to the list in the docs. I'm happy to make a PR for that if this is accepted.

### Tasks and updates

- [ ] discuss if we want to document this
- [ ] make a PR if positive

jhult:
Fixes #379 and #376.

### DIFF
--- a/docs/users/plugins.md
+++ b/docs/users/plugins.md
@@ -1,7 +1,7 @@
 # Plugins
 
-Mdformat offers an extensible plugin system for both code fence content formatting and Markdown parser extensions (like GFM tables).
-This document explains how to use plugins.
+Mdformat offers an extensible plugin system for code fence content formatting, Markdown parser extensions (like GFM tables),
+and modifying/adding other functionality. This document explains how to use plugins.
 If you want to create a new plugin, refer to the [contributing](../contributors/contributing.md) docs.
 
 ## Code formatter plugins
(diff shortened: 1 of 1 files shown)


jhult:
@hukkin, any thoughts on getting this merged?

KyleKing (approved):

hukkin:
Thanks!

PR #427, opened 2024-03-08
📚 DOCS: add `mdformat-gfm-alerts`

KyleKing:
Add link to a new plugin for [`mdformat-gfm-alerts`](https://github.com/KyleKing/mdformat-gfm-alerts)

Requires #443 to resolve CI failure

PR #440, opened 2024-08-08
Added `mdformat-gfm-alerts` to GitHub full installation in README.md

RealA10N:
The documentation declares a list of packages that should be installed to enable full GitHub-flavored support. However, the list does not include the `mdformat-gfm-alerts` packages that adds support to alerts, which are not supported by any other package in the existing list.


RealA10N:
Related to #427

KyleKing (approved):

hukkin:
Thanks!

PR #425, opened 2024-02-10
DOCS: Add mdformat-ruff

Freed-Wu:

PR #443, opened 2024-09-24
🔧 MAINTAIN: Resolve CI Failures

KyleKing:
Update the RTD configuration to include the now required `build.os` setting: https://github.com/readthedocs/readthedocs.org/blob/2f48c81e9ce162bbd34229e897db84044b00fe31/docs/user/config-file/v2.rst#buildos

And to address the pre-commit/flake8 error, we need to add an ignore rule because renaming the file to resolve the warning would be breaking (e.g. https://github.com/executablebooks/mdformat-deflist/blob/919cc85cf8c750fd74550b65fa2a2a7728ca6628/mdformat_deflist/plugin.py#L5)

Merging this PR will unblock the open documentation PRs (#425, #418, #440, #427)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,3 +4,5 @@ max-complexity = 10
 # These checks violate PEP8 so let's ignore them
 extend-ignore = E203
 extend-exclude = */site-packages/*
+per-file-ignores =
+    src/mdformat/renderer/typing.py:A005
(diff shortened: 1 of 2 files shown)


hukkin:
Thanks!

I removed some clutter and merged [here](https://github.com/executablebooks/mdformat/pull/445) with you as co-author.

I wonder why readthedocs bothers with a `version` field in their conf file if they break it regardless :smile:

Issue #50, opened 2020-10-03
Add `--fast` flag to skip Markdown validation


hukkinj1:
See https://github.com/executablebooks/mdformat/pull/49#issuecomment-702740026

Add a CLI flag for skipping Markdown validation.
I would name the flag `--fast` or `--skip-validation`.

KyleKing:
Do you have a preference for which flag to add? I may have time to submit a PR by next week?

hukkin:
I think `--unsafe` and `--skip-validation` are both alright.

I always thought if things are eventually stable for years, it may be possible to change the default value of this check to off. If that happens we need the opposite flag. Then we'd have either `--safe` or `--no-skip-validation` out of which `--safe` seems much better. But that'd translate to `safe = true/false` in TOML which with `--unsafe` would slightly break the rule where CLI opt naming directly corresponds with TOML attribute naming....

Would `--no-validate` and `validate = false/true` be horrible? :smile: Either that or just `--unsafe` and `unsafe = true/false`, and don't worry about the future. Whoever implements can decide :smile:

KyleKing:
I like `--no-validate`, but I would probably default to `validate = true` because plugins often break the AST (or maybe that's just me 😆 https://github.com/search?q=org%3AKyleKing+renders+to+different+HTML&type=issues)

hukkin:
> I would probably prefer to default to validate = true

Yeah right now the default must definitely not change. And if it ever changes it may be useful for experimental plugins to be able to keep the validation on, similar to how they can turn validation off via `CHANGES_AST` already.

hukkin:
Closed by https://github.com/hukkin/mdformat/pull/496

PR #496, opened 2024-12-17
fix(#50): add --no-validate

KyleKing:
I just need to finish adding a test

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,9 @@ If a file is not properly formatted, the exit code will be non-zero.
 
 ```console
 foo@bar:~$ mdformat --help
-usage: mdformat [-h] [--check] [--version] [--number] [--wrap {keep,no,INTEGER}]
-                [--end-of-line {lf,crlf,keep}] [--exclude PATTERN]
-                [--extensions EXTENSION] [--codeformatters LANGUAGE]
-                [paths ...]
+usage: mdformat [-h] [--check] [--no-validate] [--version] [--number]
+                [--wrap {keep,no,INTEGER}] [--end-of-line {lf,crlf,keep}]
+                [--extensions EXTENSION] [--codeformatters LANGUAGE] [paths ...]
 
 CommonMark compliant Markdown formatter
 
@@ -106,6 +105,7 @@ positional arguments:
 options:
   -h, --help            show this help message and exit
   --check               do not apply changes to files
+  --no-validate         do not validate that the rendered HTML is consistent
   --version             show program's version number and exit
   --number              apply consecutive numbering to ordered lists
   --wrap {keep,no,INTEGER}
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -143,12 +143,16 @@ def run(cli_args: Sequence[str]) -> int:  # noqa: C901
                 getattr(plugin, "CHANGES_AST", False)
                 for plugin in enabled_parserplugins.values()
             )
-            if not changes_ast and not is_md_equal(
-                original_str,
-                formatted_str,
-                options=opts,
-                extensions=enabled_parserplugins,
-                codeformatters=enabled_codeformatters,
+            if (
+                not opts["no_validate"]
+                and not changes_ast
+                and not is_md_equal(
+                    original_str,
+                    formatted_str,
+                    options=opts,
+                    extensions=enabled_parserplugins,
+                    codeformatters=enabled_codeformatters,
+                )
             ):
                 print_error(
                     f'Could not format "{path_str}".',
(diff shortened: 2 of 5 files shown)


src/mdformat/_cli.py
```diff
@@ -198,6 +202,11 @@ def make_arg_parser(
     parser.add_argument(
         "--check", action="store_true", help="do not apply changes to files"
     )
+    parser.add_argument(
+        "--no-validate",
+        action="store_true",
```
hukkin, line 207: Thanks Kyle! A drive-by comment: let's use `dest="validate"` here so we get a non-negated option in the TOML file.
```suggestion
        action="store_false",
        dest="validate",
```

src/mdformat/_conf.py
```diff
@@ -59,6 +60,9 @@ def _validate_values(opts: Mapping, conf_path: Path) -> None:  # noqa: C901
     if "end_of_line" in opts:
         if opts["end_of_line"] not in {"crlf", "lf", "keep"}:
             raise InvalidConfError(f"Invalid 'end_of_line' value in {conf_path}")
+    if "validate" in opts:
+        if not isinstance(opts["validate"], bool):
+            raise InvalidConfError(f"Invalid 'validate' value in {conf_path}")
```
KyleKing, line 65: These lines aren't being recorded as tested in coverage, but they are being exercised by the addition to `test_invalid_conf_value`

Do you have any advice for what I might be missing?

KyleKing (commented):
Other than the coverage CI failures, I think this PR is ready for review!

src/mdformat/_conf.py
hukkin: From CI:
```
 Name                                                                                                Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------------------------------------------------------------------------
/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/mdformat/_conf.py      59      0     44      1    99%   64->66
-----------------------------------------------------------------------------------------------------------------------------------------------
```

It seems it's missing branch coverage jumping from line 64 to 66. So we need a test that has _valid_ value for `validate` option in the TOML file.

README.md
```diff
@@ -93,10 +93,9 @@ If a file is not properly formatted, the exit code will be non-zero.
 
 ```console
 foo@bar:~$ mdformat --help
-usage: mdformat [-h] [--check] [--version] [--number] [--wrap {keep,no,INTEGER}]
-                [--end-of-line {lf,crlf,keep}] [--exclude PATTERN]
-                [--extensions EXTENSION] [--codeformatters LANGUAGE]
-                [paths ...]
+usage: mdformat [-h] [--check] [--no-validate] [--version] [--number]
+                [--wrap {keep,no,INTEGER}] [--end-of-line {lf,crlf,keep}]
+                [--extensions EXTENSION] [--codeformatters LANGUAGE] [paths ...]
```
hukkin, line 98: This is now missing `[--exclude PATTERN]` (which only appears on Python 3.13+).

If you don't have Python 3.13 at hand, maybe just manually add it, and word wrap at around 80 wide max.

hukkin (changes requested):

src/mdformat/_conf.py
KyleKing: Oh that makes sense!

tests/test_config_file.py
```diff
@@ -149,3 +151,19 @@ def test_empty_exclude(tmp_path, capsys):
...
     assert run((str(tmp_path),)) == 0
     assert file1_path.read_text() == FORMATTED_MARKDOWN
+
+
+def test_conf_no_validate(tmp_path):
+    file_path = tmp_path / "file.md"
+    content = "1. ordered"
+    file_path.write_text(content)
+
+    with patch("mdformat.renderer._context.get_list_marker_type", return_value="?"):
```
hukkin, line 161:
```suggestion
    with mock.patch("mdformat.renderer._context.get_list_marker_type", return_value="?"):
```

tests/test_config_file.py
```diff
@@ -1,6 +1,7 @@
 from io import StringIO
 import sys
 from unittest import mock
+from unittest.mock import patch
```
hukkin, line 4:
```suggestion
```

hukkin: This seems unnecessary seeing that we import `mock` on the line above.

src/mdformat/_cli.py
```diff
@@ -197,6 +201,13 @@ def make_arg_parser(
     parser.add_argument(
         "--check", action="store_true", help="do not apply changes to files"
     )
+    parser.add_argument(
+        "--no-validate",
+        action="store_false",
+        dest="validate",
+        default=None,
```
hukkin, line 208:
```suggestion
        action="store_const",
        const=True,
        dest="validate",
```

hukkin: End result is probably the same, but I think this may be more in line with what the other flags do.

hukkin (changes requested):
Two minor nits, otherwise LGTM!

src/mdformat/_cli.py
KyleKing: I think this needs to be `const=False`, but I like that this is more explicit

PR #504, opened 2024-12-27
fix: Thematic breaks should respect wrap length

sjoblomj:
Currently, a thematic break will always be 70 characters. This PR makes the thematic break respect the given wrap length. 

- If the user gives no wrap length, 70 characters will be used as per today.
- If the user gives a very small wrap length, 3 characters will be used, as per [the spec](https://spec.commonmark.org/0.31.2/#thematic-breaks).


hukkin:
Hey, thanks for the PR!

The --wrap feature is really only intended and documented to wrap paragraph text. This would be a new feature that can be considered in a separate issue or together with larger changes here https://github.com/hukkin/mdformat/issues/331.

KyleKing:
One other option to consider is contributing to this existing plugin: https://github.com/csala/mdformat-simple-breaks/issues/1

PR #512, opened 2025-02-01
fix(#507): warn when CLI defaults conflict with TOML

### LINKED ISSUE #507: Warning on CLI argument default
### Describe the bug

**context**

When `mdformat` or a plugin add a cli argument with a default (e.g. 'action=store_false', 'action=store_true', `def=True`), the stored default will override any settings read from the TOML file

Previously caused me issues from https://github.com/KyleKing/mdformat-mkdocs/commit/5c7be8c7fea5c6f3199d4a203a443bd724f7e1be and https://github.com/hukkin/mdformat/pull/496#discussion_r1891346960.

**bug**

When using `store_false` and running `mdformat .`, the plugin option appears as False even if found in `.mdformat.toml` because of how the dictionaries are merged

**fix**

Replacing `store_false` with `store_const` and `const=False` will remove the default and behave as expected (when no CLI argument is passed, the TOML option takes precedence.

### Reproduce the bug

1. Use `store_true`
1. Create `.mdformat.toml` with configuration option set to True
1. Run `mdformat tmp.md`
1. Configuration option was resolved as False

### List your environment

```
mdformat 0.7.21 (mdformat_frontmatter 2.0.8, mdformat-gfm 0.4.1, mdformat_simple_breaks 0.0.1, mdformat_tables 1.0.0, mdformat_mkdocs 4.1.2, mdformat_footnote 0.1.1, mdformat_wikilink 0.2.0, mdformat-config 0.2.1, mdformat-web 0.2.0, mdformat-beautysh 0.1.1, mdformat-ruff 0.1.3)   
```

### Proposal

Could `mdformat` output a warning when any CLI argument is configured with a default to prevent this footgun? The remediation is straightforward, but the root cause is non-obvious for plugin developers

As a quick proof of concept, something like:

```py
import argparse

parser = argparse.ArgumentParser()
group = parser.add_argument_group(title="group")

# Only `--correct` will work with `.mdformat.toml`
parser.add_argument("--yes", action="store_true")
group.add_argument("--no", action="store_false")
parser.add_argument("--correct", action="store_const", const=True)

# Could check all arguments and not just plugins, but I'm not sure if there is a non-private way to access them
for action in parser._actions:
    # Would want a better way to differentiate _HelpAction from others
    if action.default not in {None, "==SUPPRESS=="}:
        print("Found action with default", action)

print("Parsed:", parser.parse_args())
```

KyleKing:
Fixes #507

Adds a user-warning when plugins have argument defaults that conflict with the new TOML config

```
> uv tool install . --editable --force --with="mdformat-mkdocs[recommended]==4.1.1"
> mdformat --version
/Users/kyleking/.local/share/uv/tools/mdformat/lib/python3.13/site-packages/mdformat_mkdocs/__init__.py:0: UserWarning: The `default` (False) for ['--align-semantic-breaks-in-lists'] from the 'mkdocs' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`. To resolve, consider refactoring to `.add_argument(..., default=None)` 
/Users/kyleking/.local/share/uv/tools/mdformat/lib/python3.13/site-packages/mdformat_mkdocs/__init__.py:0: UserWarning: The `default` (False) for ['--ignore-missing-references'] from the 'mkdocs' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`. To resolve, consider refactoring to `.add_argument(..., default=None)`
mdformat 0.7.22 (mdformat_frontmatter 2.0.8, mdformat-gfm 0.4.1, mdformat_simple_breaks 0.0.1, mdformat_tables 1.0.0, mdformat_mkdocs 4.1.1, mdformat_footnote 0.1.1, mdformat_wikilink 0.2.0, mdformat-config 0.2.1, mdformat-web 0.2.0, mdformat-beautysh 0.1.1, mdformat-ruff 0.1.3)
```

**Edit**: the example STDERR above was updated to reflect the implementation changes

### DIFF
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -303,29 +303,14 @@ def make_arg_parser(
             plugin.add_cli_argument_group(group)
             for action in group._group_actions:
                 action.dest = f"plugin.{plugin_id}.{action.dest}"
-                if type(action) in {
-                    argparse._StoreTrueAction,
-                    argparse._StoreFalseAction,
-                }:
+                if not (action.default is None or action.default == argparse.SUPPRESS):
                     import warnings
 
-                    is_store_true = type(action) is argparse._StoreTrueAction
-                    plugin_name = (
-                        action.container.title
-                        if hasattr(action, "container")
-                        else str(plugin)
-                    )
-                    text = (
-                        f"For {action.option_strings} from {plugin_name},"
-                        " the default will always override a value configured"
-                        " in TOML. To resolve, replace `.add_argument(...,"
-                        f' action="store_{str(is_store_true).lower()}")` with'
-                        f' `(..., action="store_const", const={is_store_true})`'
-                    )
+                    text = f"The argument default for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`"  # noqa: E501
                     plugin_file, plugin_line = get_source_file_and_line(action)
                     warnings.warn_explicit(
                         text,
-                        UserWarning,
+                        DeprecationWarning,
                         filename=plugin_file,
                         lineno=plugin_line,
                     )
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -309,7 +309,7 @@ def make_arg_parser(
                     plugin_file, plugin_line = get_source_file_and_line(plugin)
                     warnings.warn_explicit(
                         f"The `default` ({action.default}) for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`. To resolve, consider refactoring to `.add_argument(..., default=None)` ",  # noqa: E501
-                        DeprecationWarning,
+                        UserWarning,
                         filename=plugin_file,
                         lineno=plugin_line,
                     )
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -308,8 +308,8 @@ def make_arg_parser(
 
                     plugin_file, plugin_line = get_source_file_and_line(plugin)
                     warnings.warn_explicit(
-                        f"The `default` ({action.default}) for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`. To resolve, consider refactoring to `.add_argument(..., default=None)` ",  # noqa: E501
-                        UserWarning,
+                        f"The `default` ({action.default!r}) for {action.option_strings!r} from the {plugin_id!r} plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`. To resolve, consider refactoring to `.add_argument(..., default=None)` ",  # noqa: E501
+                        DeprecationWarning,
                         filename=plugin_file,
                         lineno=plugin_line,
                     )
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -306,13 +306,7 @@ def make_arg_parser(
                 if action.default not in {None, argparse.SUPPRESS}:
                     import warnings
 
-                    text = (
-                        f"The argument default for {action.option_strings}"
-                        f" from the '{plugin_id}' plugin, will always"
-                        " override any value configured in TOML. The only"
-                        " supported CLI defaults are `None` or `argparse."
-                        "SUPPRESS`"
-                    )
+                    text = f"The argument default for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`"  # noqa: E501
                     plugin_file, plugin_line = get_source_file_and_line(action)
                     warnings.warn_explicit(
                         text,
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -307,7 +307,7 @@ def make_arg_parser(
                     import warnings
 
                     text = f"The argument default for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`"  # noqa: E501
-                    plugin_file, plugin_line = get_source_file_and_line(action)
+                    plugin_file, plugin_line = get_source_file_and_line(plugin)
                     warnings.warn_explicit(
                         text,
                         DeprecationWarning,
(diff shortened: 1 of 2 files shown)


src/mdformat/_cli.py
```diff
@@ -303,6 +303,32 @@ def make_arg_parser(
             plugin.add_cli_argument_group(group)
             for action in group._group_actions:
                 action.dest = f"plugin.{plugin_id}.{action.dest}"
+                if type(action) in {
+                    argparse._StoreTrueAction,
+                    argparse._StoreFalseAction,
+                }:
```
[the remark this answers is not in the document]
hukkin: Yes,  let's broaden this to what you suggested in the issue report:

```python
if action.default is None or action.default == argparse.SUPPRESS:
```

Including `argparse.SUPPRESS` here makes perfect sense.

src/mdformat/_cli.py
```diff
@@ -303,6 +303,32 @@ def make_arg_parser(
...
+                        f"For {action.option_strings} from {plugin_name},"
+                        " the default will always override a value configured"
+                        " in TOML. To resolve, replace `.add_argument(...,"
+                        f' action="store_{str(is_store_true).lower()}")` with'
+                        f' `(..., action="store_const", const={is_store_true})`'
+                    )
+                    plugin_file, plugin_line = get_source_file_and_line(action)
+                    warnings.warn_explicit(
+                        text,
+                        UserWarning,
```
hukkin, line 328: Let's use `DeprecationWarning` to avoid the visibility issue mentioned here https://github.com/hukkin/mdformat/issues/507#issuecomment-2612785031

src/mdformat/_cli.py
```diff
@@ -303,6 +303,32 @@ def make_arg_parser(
...
+                    argparse._StoreFalseAction,
+                }:
+                    import warnings
+
+                    is_store_true = type(action) is argparse._StoreTrueAction
+                    plugin_name = (
+                        action.container.title
+                        if hasattr(action, "container")
+                        else str(plugin)
+                    )
```
hukkin, line 317: I'd remove this and use `f"{plugin_id} plugin"` in the `text` string below.

src/mdformat/_cli.py
```diff
@@ -303,6 +303,32 @@ def make_arg_parser(
...
+                    text = (
+                        f"For {action.option_strings} from {plugin_name},"
+                        " the default will always override a value configured"
+                        " in TOML. To resolve, replace `.add_argument(...,"
+                        f' action="store_{str(is_store_true).lower()}")` with'
+                        f' `(..., action="store_const", const={is_store_true})`'
+                    )
```
hukkin, line 324: I think just recommending to add `default=None` or `default=argparse.SUPPRESS` would work regardless of the action used. Should also work with `action=store_true/false`, so I don't think there's even a need to mention them by name (I know I said otherwise in an issue comment, sorry :smile: )

hukkin (changes requested):

src/mdformat/_cli.py
```diff
@@ -303,6 +303,17 @@ def make_arg_parser(
             plugin.add_cli_argument_group(group)
             for action in group._group_actions:
                 action.dest = f"plugin.{plugin_id}.{action.dest}"
+                if action.default not in {None, argparse.SUPPRESS}:
+                    import warnings
+
+                    text = f"The argument default for {action.option_strings} from the '{plugin_id}' plugin, will always override any value configured in TOML. The only supported CLI defaults are `None` or `argparse.SUPPRESS`"  # noqa: E501
```
KyleKing, line 309: Feel free to revert the last commit before merging if you prefer the string to be hard-wrapped.

I thought black would manage strings like that automatically, but I'm not sure if there a missing config in pre-commit or a version issue?

KyleKing (commented):
Thanks for the feedback!

src/mdformat/_cli.py
hukkin: Black doesn't wrap strings, to my knowledge (I could be wrong). I don't mind the wrapping, either way works for me.

I really liked how you had concrete instructions on how to fix, in the previous edit. Could you add something similar like ```To resolve, do `.add_argument(..., default=None)` ``` ?

Maybe also move the string inside `warn_explicit` to not pollute the namespace with a generic name.

src/mdformat/_cli.py
```diff
@@ -303,6 +303,17 @@ def make_arg_parser(
...
+                    plugin_file, plugin_line = get_source_file_and_line(action)
```
hukkin, line 310:
```suggestion
                    plugin_file, plugin_line = get_source_file_and_line(plugin)
```

src/mdformat/_cli.py
KyleKing: The risk with a deprecation warning is that developers may never see it because the test code needs to exercise `run()` or the plugin author needs to modify the warning level, which when tested with my prior release of `mdformat-mkdocs` wouldn't have been visible to me

I think a `UserWarning` might be appropriate because ultimately users will be the ones experiencing unexpected behavior. The impact should be small because only plugins that have migrated to the new groups *and* have arguments with defaults will need to make any changes. For examples `mdformat-tables=1.0.0` won't error and I don't see other examples on GitHub: https://github.com/search?q=add_cli_argument_group+language%3APython+NOT+repo%3Ahukkin%2Fmdformat+NOT+repo%3Ajamesquilty%2Fmdformat+NOT+repo%3Aantazoey%2Fmdformat+NOT+repo%3Ajamesquilty%2Fmdformat-tables+NOT+repo%3Ahukkin%2Fmdformat-tables+NOT+repo%3AKyleKing%2Fmdformat&type=code&l=Python

src/mdformat/_cli.py
KyleKing: I've updated the description above (https://github.com/hukkin/mdformat/pull/512#issue-2825317428) with the new output. Feel free to edit the text directly or let me know if you have other feedback!

tests/test_plugins.py
```diff
@@ -272,6 +272,35 @@ def test_cli_options_group(monkeypatch, tmp_path):
...
+            group.add_argument("--store-false", action="store_false")
+            group.add_argument("--store-zero", default=0)
+            group.add_argument("--store-const", action="store_const", const=True)
+
+    monkeypatch.setitem(PARSER_EXTENSIONS, "table", ExamplePluginWithStoreTrue)
+    file_path = tmp_path / "test_markdown.md"
+    file_path.touch()
+
+    with patch.object(MDRenderer, "render", return_value=""):
+        with pytest.warns(UserWarning) as warnings:
```
hukkin, line 295:
```suggestion
        with pytest.warns(DeprecationWarning) as warnings:
```

hukkin (changes requested):

PR #515, opened 2025-02-08
Fix stdin encoding

chmp:
Change: Use utf8-encoding when reading files from stdin. 

The current code uses stdout.buffer.read() + encode for writing, which implies Utf8 encoding, but uses stdin.read(), which uses an environment specific encoding. See [docs](https://docs.python.org/3/library/sys.html#sys.stdin).

I ran into this issue when using Mdformat with Helix on Windows. For files with German content and Utf8-encoding the current code mangled any non-ascii characters. The proposed code fixed this behavior.

### DIFF
--- a/docs/users/changelog.md
+++ b/docs/users/changelog.md
@@ -3,10 +3,11 @@
 This log documents all Python API or CLI breaking backwards incompatible changes.
 Note that there is currently no guarantee for a stable Markdown formatting style across versions.
 
-## Develop
+## **unreleased**
 
-- Use consistent encoding when reading from stdin.
-  It is now read with UTF-8 encoding, consistent with the encoding used in writing.
+- Fixed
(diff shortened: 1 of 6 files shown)


chmp:
@hukkin oh. I just executed the command itself - I will fix the test.

chmp:
I fixed the tests using stdin. Locally, tests using symlinking are failing for me, but that's due to Windows, I guess.

hukkin:
Thanks for this!

flake8 complains:
```
tests/test_cli.py:1:1: F401 'io.StringIO' imported but unused
tests/test_config_file.py:1:1: F401 'io.StringIO' imported but unused
```
Would you able to fix this?

chmp:
@hukkin Sorry for the late reply. I did not have access to my computer for a couple of days. Thanks a lot for fixing the last issues and merging!

PR #520, opened 2025-02-24
docs: correct mdformat-admon plugin name

### LINKED ISSUE #519: KeyError: 'admonition'
### Describe the bug

**context**
When I do `mdformat.text(value, extensions={"tables", "admonition"})`.

**expectation**
I expected no error to occur.

**bug**
But instead a `KeyError` happens.
Here's an error message I ran into...

```python
import mdformat
mdformat.text("foo bar", extensions={"tables", "admonition"})
```

### Reproduce the bug

1. Install all extensions.
2.  ```python
    import mdformat
    mdformat.text("foo bar", extensions={"tables", "admonition"})
    ```

### List your environment

```
❯ mdformat --version
mdformat 0.7.22 (mdformat-gfm 0.4.1, mdformat_tables 1.0.0, mdformat_gfm_alerts 1.0.1, mdformat_admon 2.0.6)
```

KyleKing:
Fixes #519

### DIFF
--- a/docs/users/plugins.md
+++ b/docs/users/plugins.md
@@ -107,7 +107,7 @@ Explore mdformat's [GitHub topic](https://github.com/topics/mdformat) for more.
   </tr>
   <tr>
     <td><a href="https://github.com/KyleKing/mdformat-admon">mdformat-admon</a></td>
-    <td><code>admonition</code></td>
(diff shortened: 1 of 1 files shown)


hukkin:
Thank you!

PR #532, opened 2025-08-16
add: option to return non-zero if files are changed

chadrik:
Fixes https://github.com/hukkin/mdformat/issues/531


nrusch:
I'd be very interested in this feature as well.

tests/test_cli.py
```diff
@@ -39,6 +46,19 @@ def test_format__folder(tmp_path):
...
 
 
+def test_format__folder_exit_non_zero(tmp_path):
+    file_path_1 = tmp_path / "test_markdown1.md"
+    file_path_2 = tmp_path / "test_markdown2.md"
+    file_path_3 = tmp_path / "not_markdown3"
+    file_path_1.write_text(UNFORMATTED_MARKDOWN)
+    file_path_2.write_text(UNFORMATTED_MARKDOWN)
+    file_path_3.write_text(UNFORMATTED_MARKDOWN)
+    assert run((str(tmp_path),), "--exit-non-zero-on-format") == 1
```
KyleKing, line 56: I think there might be an error in this test if I understand it correctly:

```suggestion
    assert run((str(tmp_path), "--exit-non-zero-on-format")) == 0
```

src/mdformat/_cli.py
```diff
@@ -210,6 +212,11 @@ def make_arg_parser(
     parser.add_argument(
         "--check", action="store_true", help="do not apply changes to files"
     )
+    parser.add_argument(
+        "--exit-non-zero-on-format",
+        action="store_true",
+        help="return non-zero if files are modified",
+    )
```
KyleKing, line 219: This new option should be mutually exclusive with --check (and maybe better document that both are CLI only because it doesn't make sense to configure these permanently in toml)

PR #534, opened 2025-08-29
Support markdown-it-py 4

gaborbernat:
Resolves https://github.com/hukkin/mdformat/issues/533

Ran the CI here https://github.com/gaborbernat/mdformat/actions/runs/17328552712?pr=1 and seems to be working fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ authors = [
 license = { file = "LICENSE" }
 requires-python = ">=3.9"
 dependencies = [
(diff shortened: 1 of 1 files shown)


hukkin:
Thanks!

hukkin:
See https://github.com/hukkin/mdformat/issues/545#issuecomment-3311743806

I want to do this ASAP, but a little more than a push of a button is required this time so I haven't found the time yet.

PR #563, opened 2025-11-04
Add --config option to specify explicit configuration path

meek2100:
This pull request introduces the `--config <path>` command-line option, allowing users to explicitly define the path to their TOML configuration file, overriding the default recursive search for `.mdformat.toml`.

This feature primarily supports integration with modern tooling like pre-commit hooks and centralized configuration management systems where config files may live outside the project root or in a custom directory.

### Key Changes:

1.  **New CLI Argument:** Added `--config` (type `Path`) to `src/mdformat/_cli.py`.
2.  **Config Logic:** Implemented `read_single_config_file` in `src/mdformat/_conf.py` to support direct path loading. The `run` function logic prioritizes `--config` and correctly handles non-existent file paths by exiting with an error.
3.  **Tests:** Added a new test case (`test_config_override_precedence`) verifying that the explicit `--config` correctly overrides auto-detected `.mdformat.toml` settings.
4.  **Documentation:** Updated `docs/users/configuration_file.md` to document the new usage.
5.  **Code Quality Fix:** Corrected the `InvalidPath` exception class in `src/mdformat/_cli.py` to properly call `super().__init__(path)`, resolving the `flake8-bugbear` (B042) warning.

### DIFF
--- a/src/mdformat/_cli.py
+++ b/src/mdformat/_cli.py
@@ -12,7 +12,7 @@ import sys
 import textwrap
 
 import mdformat
-from mdformat._conf import DEFAULT_OPTS, InvalidConfError, read_toml_opts
+from mdformat._conf import DEFAULT_OPTS, InvalidConfError, read_toml_opts, read_single_config_file,
 from mdformat._util import detect_newline_type, is_md_equal
 import mdformat.plugins
 
@@ -34,6 +34,10 @@ def run(cli_args: Sequence[str], cache_toml: bool = True) -> int:  # noqa: C901
     }
     cli_core_opts, cli_plugin_opts = separate_core_and_plugin_opts(cli_opts)
 
+    config_override_path = cli_core_opts.pop("config", None)
+    if config_override_path and not isinstance(config_override_path, Path):
+        config_override_path = Path(config_override_path)
+
     if not cli_opts["paths"]:
         print_paragraphs(["No files have been passed in. Doing nothing."])
         return 0
@@ -46,12 +50,20 @@ def run(cli_args: Sequence[str], cache_toml: bool = True) -> int:  # noqa: C901
     format_errors_found = False
     renderer_warning_printer = RendererWarningPrinter()
     for path in file_paths:
-        read_toml = read_toml_opts if cache_toml else read_toml_opts.__wrapped__
         try:
-            toml_opts, toml_path = read_toml(path.parent if path else Path.cwd())
+            if config_override_path:
+                toml_opts, toml_path = read_single_config_file(config_override_path)
+            else:
+                read_toml = read_toml_opts if cache_toml else read_toml_opts.__wrapped__
+                toml_opts, toml_path = read_toml(path.parent if path else Path.cwd())
         except InvalidConfError as e:
             print_error(str(e))
             return 1
+        except FileNotFoundError as e:
+            if config_override_path and config_override_path.samefile(Path(e.filename)):
+                print_error(f"Configuration file not found at: {e.filename}")
+                return 1
+            raise
 
         opts = {**DEFAULT_OPTS, **toml_opts, **cli_core_opts}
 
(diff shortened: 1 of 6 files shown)


meek2100:
@hukkin,
This should be ready to go and would be very helpful to have merged for one of my projects.

src/mdformat/_conf.py
KyleKing: Maybe this snippet could be extracted into a shared helper? The only difference is the error message (which includes the config_path in the former)

src/mdformat/_conf.py
```diff
@@ -32,6 +32,23 @@ class InvalidConfError(Exception):
     """
 
 
+def read_single_config_file(config_path: Path) -> tuple[Mapping, Path | None]:
```
KyleKing: ~~This function should be cached like read_toml_opts~~ (not strictly necessary, see proposal to hoist outside of the for loop)

src/mdformat/_cli.py
```diff
@@ -46,12 +55,20 @@ def run(cli_args: Sequence[str], cache_toml: bool = True) -> int:  # noqa: C901
...
+                read_toml = read_toml_opts if cache_toml else read_toml_opts.__wrapped__
+                toml_opts, toml_path = read_toml(path.parent if path else Path.cwd())
         except InvalidConfError as e:
             print_error(str(e))
             return 1
+        except FileNotFoundError as e:
+            if config_override_path and str(config_override_path) == str(e.args[0]):
+                print_error(f"Configuration file not found at: {e.args[0]}")
+                return 1
+            raise
```
KyleKing: What about raising a custom ConfigFileNotFoundError or ConfigOverrideNotFoundError (instead of the more general FileNotFoundError) with the relevant metadata, then this conditional logic isn't necessary?

src/mdformat/_cli.py
```diff
@@ -46,12 +55,20 @@ def run(cli_args: Sequence[str], cache_toml: bool = True) -> int:  # noqa: C901
     format_errors_found = False
     renderer_warning_printer = RendererWarningPrinter()
     for path in file_paths:
-        read_toml = read_toml_opts if cache_toml else read_toml_opts.__wrapped__
         try:
-            toml_opts, toml_path = read_toml(path.parent if path else Path.cwd())
+            if config_override_path:
+                toml_opts, toml_path = read_single_config_file(config_override_path)
```
KyleKing: Actually on caching, the config override file can be read *outside* of the loop because the configuration file isn't relative to the file path. Maybe something like:

```python
(toml_opts, toml_path): tuple[...?, Path] = None, None
if config_override_path:
    toml_opts, toml_path = read_single_config_file(config_override_path)
    for path in file_paths:
        ...
```

KyleKing (approved):
I added a few minor comments as an active in the mdformat community and to possibly help this PR along, but I'm not a maintainer, so take my feedback with a grain of salt!

BobbyMakelve:
Okay, so you're interested in how cryptocurrencies are created! That's a fundamental part of the crypto world.

There are primarily two main ways new cryptocurrencies are "created" or brought into existence:
1. Mining (Proof of Work): ◦   This is how Bitcoin and many other older cryptocurrencies are created. ◦   Powerful computers (miners) compete to solve complex mathematical puzzles. ◦   The first miner to solve the puzzle gets to add a new block of transactions to the blockchain. ◦   As a reward for their work, they receive a certain amount of newly minted cryptocurrency. ◦   This process secures the network and verifies transactions.  2. Staking (Proof of Stake): ◦   Newer cryptocurrencies, like Ethereum 2.0, use this method. ◦   Instead of computational power, participants "stake" (lock up) their existing cryptocurrency as collateral. ◦   The protocol selects a staker to create a new block based on how much they've staked and how long it's been staked. ◦   If they create a valid block, they earn a reward in newly minted currency. ◦   This method is generally considered more energy-efficient than mining. 
Beyond these two main methods, new coins and tokens can also be "created" through:
•   Initial Coin Offerings (ICOs) or Token Sales: Projects raise funds by selling a pre-mined or pre-generated supply of their new tokens to early investors. •   Airdrops: New tokens are distributed for free to existing wallet addresses, often as a marketing strategy to gain attention and users. •   Farming/Liquidity Mining: Users deposit existing tokens into liquidity pools on decentralized exchanges and earn new tokens as a reward for providing liquidity. 
Which aspect of creation were you thinking about specifically? Or perhaps something else entirely?

